PR #1, opened 2017-07-30
Add more tags

fengalin:
Album covers can be found in `image` as a `Sample`
Some containers (e.g. mkv) may use `album-artist` instead of `artist`


sdroege:
Thanks, merged! And as this is the first pull request, let me know if we ever meet and you'll get a beer or other drink :)

fengalin:
:) sure will!

Issue #2, opened 2017-08-02
Proper error handling in examples


sdroege:
All the unwrap()s are ugly and we can easily do better.

Idea would be to have `main` be a function that just calls `run` or a couple of other functions returning a `Result<_,_>` and it's only job is to destructure that and print any possible errors.

For each example we would then define an error enum (that implements `Display`, `Error` and what else) like
```Rust
enum MyLittleError {
    MissingElement(&str),
    LinkingFailed(&str, &str),
    Other(&str),
}
```
and return those from the functions, ideally making use of the ?-operator and various `Option<_>` / `Result<_, _>` combinators, e.g. `gst::ElementFactory::make("appsrc", None).ok_or(MyLittleError::MissingElement("appsrc"))?;`.

philn:
I started a patch :)

sdroege:
This should be simplified a lot by https://github.com/sdroege/gstreamer-rs/issues/44

sdroege:
This is now tracked at https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/issues/2

PR #16, opened 2017-08-03
examples: appsink: improved error handling

philn:
The program should now exit gracefully with human readable messages.

See also #2  :)


sdroege:
Merged, thanks!

sdroege:
It could make sense to move parts of that into macros. Element creation for example. To reduce the boilerplate a bit

PR #17, opened 2017-08-04
Add gtkplayer example

tuna74:
As discussed on IRC, I made a small GTK+ based player where the user can choose the video file. I used the code in playbin.rs and gtksink.rs as inspiration. Sound works.

If you would like any features added please tell me.

Best regards
Andreas Tunek (tuna74)

### DIFF
--- a/examples/src/bin/gtkplayer.rs
+++ b/examples/src/bin/gtkplayer.rs
@@ -10,6 +10,11 @@ use gtk::{Window, WindowType};
 
 pub mod utils;
 
+fn create_playbin() -> Result<gst::Pipeline, utils::ExampleError> {
+    let playbin = utils::create_element("playbin")?;
+    Ok(playbin.downcast::<gst::Pipeline>().unwrap())
+}
+
 fn main_loop() -> Result<(), utils::ExampleError> {
     gst::init().map_err(utils::ExampleError::InitFailed)?;
     gtk::init().map_err(utils::ExampleError::GtkInitFailed)?;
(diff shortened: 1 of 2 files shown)


sdroege:
Thanks for the PR :) Take a look at the appsink and appsrc examples, it would be nice if you could move it to some kind of better error checking than `unwrap` too

sdroege:
Also moving it over to gtk::Application would be nice. See the latest version of the gtksink.rs example

tuna74:
I'll skip GtkApplication if that is ok, it seems really messy if you have to implement the SendCell stuff.

sdroege:
Yeah, don't do that for now. We'll move that to a central place and then it will all be simple :)

tuna74:
I could not do better errors for set_state since playbin is an Element instead of a Pipeline, and I could not do better errors for set_prop and get_prop either. Maybe you could do some nice stuff for those two?

sdroege:
I'll try to take a another look later or tomorrow, I didn't forget about the PR :) I'll also check what can be improved there with the errors and give some suggestions / code. The GApplication part I can implement, that's mostly copy&paste from the other one.

Every Element is a Pipeline though, so that shouldn't be a problem (and playbin is actually a pipeline, Rust just does not know until you dynamic_cast it).

Maybe you could also consider add a seek slider? There is a simple example with GTK here (in C): https://cgit.freedesktop.org/gstreamer/gst-docs/tree/examples/tutorials/basic-tutorial-5.c

tuna74:
I'll add a seek slider and the playbin stuff.

tuna74:
I can't do 
utils::set_state(&playbin_clone, gst::State::Playing)?;

in my gtk callbacks. Do you have any good idea how to solve that?

sdroege:
Yes, in the utils it should be e.g.
```Rust
pub fn set_state<T: IsA<gst::Element>>(element: &T, state: gst::State) -> ...
```

sdroege:
You could also move this to GstPlayer now, for a more feature-full and simpler player

sdroege:
Something like this is now in the basic-tutorial-5. Let's close this for now until someone wants to revive it for writing a usable player app.

Issue #10, opened 2017-08-03
Add bindings for GstToc/TocSetter


sdroege:
https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstToc.html
https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/gstreamer-GstTocSetter.html

The TocSetter should be possible to autogenerate, Toc needs manual code (because miniobject). See e.g. caps.rs, tags.rs, sample.rs.

Needs also addition of support for the event and message

fengalin:
I will work on this one if that's ok with you.

sdroege:
Sure, go ahead :)

PR #22, opened 2017-08-09
Add bindings for GstToc/TocSetter

fengalin:
See #10


gstreamer/src/event.rs
```diff
@@ -468,8 +468,18 @@ impl<'a> StreamGroupDone<'a> {
...
-// TODO
 pub struct Toc<'a>(&'a EventRef);
+impl<'a> Toc<'a> {
+    pub fn get_toc(&self) -> (&'a ::TocRef, bool) {
+        unsafe {
+            let mut toc = ptr::null_mut();
+            let mut updated = mem::uninitialized();
+
+            ffi::gst_event_parse_toc(self.0.as_mut_ptr(), &mut toc, &mut updated);
+            (::TocRef::from_ptr(toc), updated != 0)
```
sdroege: from_glib(updated)

gstreamer/src/event.rs
```diff
@@ -932,26 +942,25 @@ impl EosBuilder {
...
     seqnum: Option<u32>,
     running_time_offset: Option<i64>,
-    #[allow(unused)]
-    toc: (),
+    toc: Option<::Toc>,
     updated: bool,
 }
 impl TocBuilder {
-    pub fn new(toc: (), updated: bool) -> Self {
+    pub fn new(toc: ::Toc, updated: bool) -> Self {
```
sdroege: Could also take a &Toc as it's to_glib_none() anyway, but not too important

gstreamer/src/lib.rs
```diff
@@ -100,6 +100,9 @@ pub use value::*;
 mod segment;
 pub use segment::*;
 
+pub mod toc;
+pub use toc::*;
```
sdroege: Instead of * only Toc and TocRef IMHO

gstreamer/src/message.rs
```diff
@@ -878,7 +878,14 @@ impl<'a> Progress<'a> {
 
 pub struct Toc<'a>(&'a MessageRef);
 impl<'a> Toc<'a> {
-    // TODO get_toc()
+    pub fn get_toc(&self) -> (::Toc, bool) {
+        unsafe {
+            let mut toc = ptr::null_mut();
+            let mut updated = mem::uninitialized();
+            ffi::gst_message_parse_toc(self.0.as_mut_ptr(), &mut toc, &mut updated);
+            (from_glib_full(toc), updated != 0)
```
sdroege: from_glib(updated)

gstreamer/src/toc.rs
```diff
@@ -0,0 +1,350 @@
+use std::ffi::CStr;
```
sdroege: Copyright header blurb missing

gstreamer/src/lib.rs
sdroege: And TocEntry / TocEntryRef

gstreamer/src/toc.rs
```diff
@@ -0,0 +1,350 @@
...
+        unsafe {
+            let toc_entry = ffi::gst_toc_find_entry(
+                self.as_ptr(),
+                uid.to_glib_none().0,
+            );
+            if toc_entry.is_null() {
+                return None;
+            }
+
+            Some(TocEntry::from_glib_borrow(
```
sdroege: This is unsafe. Either return a &TocEntryRef or from_glib_none()

from_glib_borrow() return values must only be used in this very scope (and not be passed outside) and only references be passed to functions called from this scope

gstreamer/src/toc.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    pub fn get_tags(&self) -> Option<TagList> {
+        unsafe {
+            let tags = ffi::gst_toc_get_tags(self.as_ptr());
+            if tags.is_null() {
+                return None;
+            }
+
+            Some(TagList::from_glib_borrow(tags as *const ffi::GstTagList))
```
sdroege: Same here

gstreamer/src/toc.rs
```diff
@@ -0,0 +1,350 @@
...
+            from_glib_full(ffi::gst_mini_object_copy(self.as_ptr() as *const _) as
+                *mut _)
+        }
+    }
+}
+
+unsafe impl Sync for TocRef {}
+unsafe impl Send for TocRef {}
+
+
```
sdroege: Double newline

gstreamer/src/toc.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    pub fn get_parent(&self) -> Option<TocEntry> {
+        unsafe {
+            let parent = ffi::gst_toc_entry_get_parent(self.as_mut_ptr());
+            if parent.is_null() {
+                return None;
+            }
+
+            Some(TocEntry::from_glib_borrow(
```
sdroege: No borrow

gstreamer/src/toc.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub fn get_start_stop_times(&self) -> Option<(i64, i64)> {
+        unsafe {
+            let mut start = mem::uninitialized();
+            let mut stop = mem::uninitialized();
+
+            if ffi::gst_toc_entry_get_start_stop_times(
+                self.as_ptr(),
+                &mut start,
+                &mut stop
+            ) != 0
```
sdroege: from_glib() for the bool conversion

gstreamer/src/toc.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    pub fn get_tags(&self) -> Option<TagList> {
+        unsafe {
+            let tags = ffi::gst_toc_entry_get_tags(self.as_ptr());
+            if tags.is_null() {
+                return None;
+            }
+
+            Some(TagList::from_glib_borrow(tags as *const ffi::GstTagList))
```
sdroege: No borrow

sdroege:
Generally looks good, just a few small things :) Thanks! What file are you testing that with currently?

fengalin:
Thank you for your review and comments! I think I addressed all of them.
I used a couple mkv files and flac songs from an album that comes with an m3u playlist.

fengalin:
I'm [building an application](https://github.com/fengalin/media-toc) to edit or generate toc from media files.
I gave a try to ffmpeg Rust binding and now completed the same with gstreamer.

sdroege:
I shortly looked at your application, and https://github.com/fengalin/media-toc/blob/gstreamer/src/media/audio_buffer.rs looks like it could make use of GstAudioInfo (which I'm currently working on), and https://github.com/fengalin/media-toc/blob/gstreamer/src/media/media_info.rs of GstDiscoverer (which is also somewhere on my list but should be relatively easy to do)

I'll review your changes in a bit, thanks for the fast update :)

fengalin:
Yes, I saw your work on `GstAudioInfo`. Will use it for sure.
Thanks for pointing me to `GstDiscover`.

gstreamer/src/event.rs
```diff
@@ -942,25 +942,24 @@ impl EosBuilder {
...
-            toc: Some(toc),
+            toc: toc,
             updated: updated,
         }
     }
 
     event_builder_generic_impl!(|s: &mut Self| {
-        let toc = s.toc.take();
-        ffi::gst_event_new_toc(toc.to_glib_none().0, s.updated.to_glib())
+        ffi::gst_event_new_toc(s.toc.to_glib_none().0, s.updated.to_glib())
```
sdroege: &mut Self not needed here, can be just &Self now

gstreamer/src/toc.rs
```diff
@@ -214,7 +221,7 @@ impl TocEntryRef {
                 return None;
             }
 
-            Some(TagList::from_glib_borrow(tags as *const ffi::GstTagList))
+            Some(TagList::from_glib_none(tags as *const ffi::GstTagList))
```
sdroege: FWIW, this is exactly why the Ref types for the miniobjects exist and you can borrow them. But this seems not needed here and more complicated than anything else.

sdroege:
Looks good now except for the additional &mut

gstreamer/src/event.rs
fengalin: Yes! Sorry about that.

sdroege: No problem, easy to forget these little things :)

gstreamer/src/toc.rs
fengalin: Do you mean something like this?
``` Rust
    pub fn get_tags(&self) -> Option<&TagListRef> {
        unsafe {
            let tags = ffi::gst_toc_get_tags(self.as_ptr());
            if tags.is_null() {
                return None;
            }

            Some(TagListRef::from_ptr(tags as *const ffi::GstTagList))
        }
    }
```
It seems weird to return `&TagListRef`.

sdroege: Yes, that's why I said it seems unneeded :) It will work, but just keep it as-is for now.

fengalin:
Just commited the `&mut` part. I can still work on the other issue in order to get it right.

gstreamer/src/toc.rs
fengalin: All right!

fengalin:
Add the last one if you deem it to be useful :)

sdroege:
Thanks, merged :)

sdroege:
AudioInfo bindings are there now, let me know if that makes your life a bit easier or if something else is missing :)

fengalin:
Thank you! That will be much better than my "naive" implementation. I will also use `AudioFormatInfo` `unpack` function in order to replace my own sample conversions to `f64`.

sdroege:
Unpacking will go to S32 for integer input formats though. You either want GstAudioConverter, or you could just let an audioconvert element at the end of your pipeline convert to f64. For that you could here (https://github.com/fengalin/media-toc/blob/be11572f21134efe796c87804ab1d2ce11ba11c4/src/media/context.rs#L135) have a pipeline with a tee, one branch going to an autoaudiosink like now, another going to an audioconvert and appsink for collecting the frames and f64 conversion.

fengalin:
Yes, I thought about a tee too. Integer samples would be good enough though. I only need a predictable range and 0 offset in order to [draw the waveform](https://github.com/fengalin/media-toc/blob/6aa9ad106603a0990a87b38a3921408998bb515b/src/ui/audio_controller.rs#L76).

I though about it and now my concern is to get a good balance between the time spent in GStreamer's thread in the pad probe callback and the time spent in the drawing area `draw` event. Given that the `draw` event will be executed more often (every redraw), I tend to think I need to reduce the computation needed there. That's the reason why I [uninterleaved the channels and normalized the samples](https://github.com/fengalin/media-toc/blob/6aa9ad106603a0990a87b38a3921408998bb515b/src/media/audio_buffer.rs#L161) - well I got to the point that I didn't need to interleave channels actually since seeking in an array is in o(1) -. On the other end, I don't want to hang the GStreamer thread too long and cause lags. I suspect that with current hardware, that should not happen anyway. This is interesting, I'd like to find a good operating point, but I guess it is really system dependent.

sdroege:
That's also why I suggested the tee. Basically what you would have is:
```
          --- queue --- audioconvert --- appsink
---- tee /
         \
          --- queue --- audioconvert --- audioresample --- autoaudiosink
```

The thread where you get the audio buffers in the appsink is a different one than the one that goes to the audiosink (the queue is a thread barrier), so you can spend more time doing things there without interrupting playback (as long as everything still runs in real-time on average). You should be able to completely prepare the samples there to whatever you need for the rendering in the UI thread later, so that the UI thread only has to render and not do any calculations anymore that are unrelated to the visualization.

You would then get e.g. F64 in the appsink, could deinterleave there as needed, normalize samples, etc and just have the amplitudes of the waveform you want to draw as a result there.

fengalin:
Great! I'll do that. Many thanks for the wise advices.

sdroege:
https://github.com/sdroege/gstreamer-rs/commit/b655c838b285b4dea8dff4f36a2c340b092b8b66 is probably also useful for you, to simplify things even further

fengalin:
Yes indeed! Thanks again!

sdroege:
API might still minimally change, but that should be no problem to use it :)

fengalin:
No problem! I'll give you a feedback soon.

sdroege:
Great thanks, let me know if something could be nicer

PR #27, opened 2017-08-15
Define Sync and Send traits for AudioInfo...

fengalin:
... if this is acceptable.

I need this in my `sink_pad.add_probe` closure, because I'd like to avoid reading `AudioInfo` (and `Caps`) on every buffer reception. So I defined a `Mutex<Option<AudioInfo>>` initialized to `None` outside of the closure. When the first buffer arrives, I read the `AudioInfo`, set the `Option` to `Some(audio_info)` and reuse it for subsequent buffers.

Full code is [here](https://github.com/fengalin/media-toc/blob/400649d9a768b7e3138ad4da777cc63dc22e59f7/src/media/context.rs#L172).

### DIFF
--- a/gstreamer-audio/src/audio_info.rs
+++ b/gstreamer-audio/src/audio_info.rs
@@ -263,6 +263,9 @@ impl PartialEq for AudioInfo {
 
 impl Eq for AudioInfo {}
 
+unsafe impl Sync for AudioInfo {}
+unsafe impl Send for AudioInfo {}
+
 impl glib::types::StaticType for AudioInfo {
     fn static_type() -> glib::types::Type {
         unsafe { glib::translate::from_glib(ffi::gst_audio_info_get_type()) }


sdroege:
Oh that's fine of course, thanks! AudioInfo (in Rust) is immutable. In C it isn't but that doesn't matter for us here.

sdroege:
For your code, the caps can also change. It would be better to either use the notify::caps signal on the pad, or also collect caps events from the pad probe. All buffers coming after a caps change are going to be according to the new caps.

sdroege:
Also I think creating gtk(gl)sink from the pad probe (from some arbitrary streaming thread) and also getting the widget there is not going to work well. I think that one has to be created (and set to PLAYING) on the GTK main thread.

fengalin:
- Caps/probe: Thanks! I'll do that. Is this something which can only happen on a streamed media, or is it also possible while reading a file?
- GTK(GL)Sink: I'll have to check again. I tried using a widget from the GTK thread, but it didn't work. The option in the code right now is the only one that I managed to get to work properly. I would need to re-check though as I was fighting with several things at that moment. Anyway, It works ok right now :)

fengalin:
FYI, I didn't have time to try the tee for the audio playback / waveform pre-processing fork, but it's in my TODO list.

sdroege:
It can in theory happen all the time (caps changes), but unlikely either way. Maybe just do that together with switching to the tee/appsink solution (it will be much simpler than what you have right now btw) :)

For gtk(gl)sink, if it works, it works. But it is probably racy and will not work on Windows :) The better solution would be to create it in the UI thread. Maybe by blocking the video pad instead, sending the pad to the UI thread, then connecting it to a newly create sinks, then unblocking the video pad once the linking is finished.

fengalin:
I see. Note however that I add the widget to the UI in the GTK thread though and block the sink pad until it's added. I think this is pretty close to the examples I could find in the GStreamer howtos on the subject. Will try your solution to understand the details.
Thanks again!

sdroege:
OK, instead of using a channel for waiting in the pad probe you can also use a blocking pad probe. If you return OK from it, the pad will be blocked until you remove the pad probe by the ID you got when adding it.

fengalin:
I think I need to try it to figure out the implications. I'll report back!

Issue #32, opened 2017-09-08
Port all GStreamer tutorials


sdroege:
Code can be found here: https://cgit.freedesktop.org/gstreamer/gst-docs/tree/examples/tutorials
Text with explanations here: https://gstreamer.freedesktop.org/documentation/tutorials/index.html

Would be nice to port these to Rust

sdroege:
Start of this can be found here: https://github.com/sdroege/gstreamer-rs/tree/master/tutorials

Write here if someone wants to work on porting a specific tutorial

Jouan:
I can try porting basic-tutorial-2.

sdroege:
> I can try porting basic-tutorial-2.

That would be great, thanks! Let me know if you have questions

sdroege:
basic-tutorial-2 is also done. Someone wants to do basic-tutorial-3? There's code in the examples that is almost like this already (decodebin.rs)

sireliah:
Hey, I can try to write the third one. I'll use previous examples as a base.

sdroege:
Thanks, that would be great :)

sdroege:
Tutorial 3 is available here now: https://github.com/sdroege/gstreamer-rs/pull/38

lovebug356:
I also tried basic-tutorial-4. The query part works, but it is not clear how to write GST_SEEK_FLAG_FLUSH | GST_SEEK_FLAG_KEY_UNIT in rust?

https://github.com/lovebug356/gstreamer-rs/commit/0c9d36d537f89069448b8e9871740015d6d6d704

sdroege:
@lovebug356 This should work `gst::SeekFlags::FLUSH | gst::SeekFlags::KEY_UNIT` (if you use master, which uses bitflags 1.0. With 0.9 it was `gst::SEEK_FLAG_FLUSH` IIRC).

lovebug356:
Fyi, I'm working on tutorial-6, but I'm struggling a bit with adding bindings for StaticPadTemplate (gir does not build on my system due to [a problem](https://github.com/alexcrichton/git2-rs/issues/263) with git2-rs).

raytiley:
Figured I would comment on this issue rather than open a new one. Just looking into gstreamer, and have been wanting to a reason to learn a bit more about rust, so I figured I would check this project out.

I cloned the repo and ran `cargo build` inside the `tutorials` directory and received errors:

```
error: associated constants are experimental (see issue #29646)
[...]

PR #33, opened 2017-09-12
Added basic-tutorial-2

Jouan:
Contributes to solve #32.

I have never used gstreamer-rs before, so I'm not entirely sure if this is correct.
It compiles, runs and prints "Unexpected message received."

### DIFF
--- a/tutorials/src/bin/basic-tutorial-2.rs
+++ b/tutorials/src/bin/basic-tutorial-2.rs
@@ -1,0 +1,59 @@
+extern crate gstreamer as gst;
+
+use gst::GstObjectExt;
+use gst::ElementExt;
+use gst::MessageView;
+
+fn main() {
+    // Initialize GStreamer
+    gst::init().unwrap();
+
+    if let (Some(source), Some(sink)) = (
+    // Create the elements
+    gst::ElementFactory::make("videotestsrc", "source"),
+    gst::ElementFactory::make("autovideosink", "sink")) {
+        // Create the empty pipeline
+        let pipeline = gst::Pipeline::new("test-pipeline");
+
+        // Build the pipeline
+        gst::BinExtManual::add_many(
+            &pipeline,
+            &[&source, &sink]).unwrap();
+        if let Err(_) = gst::ElementExt::link(&source, &sink) {
+            eprintln!("Elements could not be linked.");
+            ::std::process::exit(-1);
+        }
+
+        // Modify the source's properties
+        gst::GObjectExtManualGst::set_property_from_str(
+            &source,
+            "pattern",
+            "0");
+
+        // Start playing
+        let ret = pipeline.set_state(gst::State::Playing);
+        if ret == gst::StateChangeReturn::Failure {
+            eprintln!("Unable to set the pipeline to the playing state.");
+            ::std::process::exit(-1);
+        }
+
+        // Wait until error or EOS
+        let bus = pipeline.get_bus().unwrap();
+        if let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
+            match msg.view() {
+                MessageView::Error(err) => {
+                    eprintln!("Error received from element {}: {}",
+                             msg.get_src().get_path_string(), err.get_error());
+                    eprintln!("Debugging information: {:?}", err.get_debug());
+                },
+                MessageView::Eos(..) => println!("End-Of-Stream reached."),
+                // We should not reach here because we only asked for ERRORs and EOS
+                _ => eprintln!("Unexpected message received."),
+            }
+        }
+    }
+    else {
+        println!("Not all elements could be created.");
+        ::std::process::exit(-1);
+    }
+}
--- a/tutorials/src/bin/basic-tutorial-2.rs
+++ b/tutorials/src/bin/basic-tutorial-2.rs
@@ -29,7 +29,7 @@ fn main() {
 
     // Wait until error or EOS
     let bus = pipeline.get_bus().unwrap();
-    if let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
+    while let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
         use gst::MessageView;
         match msg.view() {
             MessageView::Error(err) => {
@@ -37,8 +37,7 @@ fn main() {
                          msg.get_src().get_path_string(), err.get_error());
                 eprintln!("Debugging information: {:?}", err.get_debug());
             },
-            MessageView::Eos(..) => println!("End-Of-Stream reached."),
-            // We should not reach here because we only asked for ERRORs and EOS
+            MessageView::Eos(..) => break,
             _ => (),
         }
     }
--- a/tutorials/src/bin/basic-tutorial-2.rs
+++ b/tutorials/src/bin/basic-tutorial-2.rs
@@ -1,59 +1,45 @@
 extern crate gstreamer as gst;
 
-use gst::GstObjectExt;
-use gst::ElementExt;
-use gst::MessageView;
+use gst::prelude::*;
 
 fn main() {
     // Initialize GStreamer
     gst::init().unwrap();
 
-    if let (Some(source), Some(sink)) = (
     // Create the elements
-    gst::ElementFactory::make("videotestsrc", "source"),
-    gst::ElementFactory::make("autovideosink", "sink")) {
-        // Create the empty pipeline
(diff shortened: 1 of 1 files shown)


tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,59 @@
+extern crate gstreamer as gst;
+
+use gst::GstObjectExt;
+use gst::ElementExt;
+use gst::MessageView;
+
```
sdroege: Ideally you would "use gst;" and "use gst::prelude::*;" here

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,59 @@
...
+                    eprintln!("Debugging information: {:?}", err.get_debug());
+                },
+                MessageView::Eos(..) => println!("End-Of-Stream reached."),
+                // We should not reach here because we only asked for ERRORs and EOS
+                _ => eprintln!("Unexpected message received."),
+            }
+        }
+    }
+    else {
+        println!("Not all elements could be created.");
```
sdroege: This else case could be at the very top and an early return from main(). Create both elements, check if both could be created and if not just error out

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,59 @@
...
+        if let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
+            match msg.view() {
+                MessageView::Error(err) => {
+                    eprintln!("Error received from element {}: {}",
+                             msg.get_src().get_path_string(), err.get_error());
+                    eprintln!("Debugging information: {:?}", err.get_debug());
+                },
+                MessageView::Eos(..) => println!("End-Of-Stream reached."),
+                // We should not reach here because we only asked for ERRORs and EOS
+                _ => eprintln!("Unexpected message received."),
```
sdroege: There is no filtered API for GstBus in the Rust bindings currently, so you would just ignore all other messages here

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,59 @@
...
+        let ret = pipeline.set_state(gst::State::Playing);
+        if ret == gst::StateChangeReturn::Failure {
+            eprintln!("Unable to set the pipeline to the playing state.");
+            ::std::process::exit(-1);
+        }
+
+        // Wait until error or EOS
+        let bus = pipeline.get_bus().unwrap();
+        if let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
+            match msg.view() {
```
sdroege: Here I would "use gst::MessageView;". Check the other examples

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,59 @@
...
+        if let Err(_) = gst::ElementExt::link(&source, &sink) {
+            eprintln!("Elements could not be linked.");
+            ::std::process::exit(-1);
+        }
+
+        // Modify the source's properties
+        gst::GObjectExtManualGst::set_property_from_str(
+            &source,
+            "pattern",
+            "0");
```
sdroege: You can even give the pattern a name here: 0 is "smpte"

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,59 @@
...
+    gst::ElementFactory::make("videotestsrc", "source"),
+    gst::ElementFactory::make("autovideosink", "sink")) {
+        // Create the empty pipeline
+        let pipeline = gst::Pipeline::new("test-pipeline");
+
+        // Build the pipeline
+        gst::BinExtManual::add_many(
+            &pipeline,
+            &[&source, &sink]).unwrap();
+        if let Err(_) = gst::ElementExt::link(&source, &sink) {
```
sdroege: `source.link(&sink)` or `gst::Element::link_many(&source, &sink)`

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,59 @@
...
+    // Create the elements
+    gst::ElementFactory::make("videotestsrc", "source"),
+    gst::ElementFactory::make("autovideosink", "sink")) {
+        // Create the empty pipeline
+        let pipeline = gst::Pipeline::new("test-pipeline");
+
+        // Build the pipeline
+        gst::BinExtManual::add_many(
+            &pipeline,
+            &[&source, &sink]).unwrap();
```
sdroege: `pipeline.add_many(&[&source, &sink])`

tutorials/src/bin/basic-tutorial-2.rs
sdroege: Also on error here, you could as well `expect("Failed to link elements)` instead of manually unwrapping the `Result` and exiting the process (same in the other places: element creation for example).

tutorials/src/bin/basic-tutorial-2.rs
Jouan: This part is mimicking: g_object_set (source, "pattern", 0, NULL);

My understanding was that "pattern" is the property to set, and "0" is the value for that property.

Could you explain what you mean by "give the pattern a name"?

sdroege: The equivalent to the C code would be with `set_property` but we can do better here. Either you could use that in combination with `glib::EnumClass` (see playbin.rs for an example with flags, enums work very similar). That would be most explicit. Or you could use `source.set_property_from_str("pattern", "smpte") `

The advantage with using the string is that it is more descriptive. You don't have to look up in the docs that 0 is the SMPTE pattern but the code just says so. And you could replace it directly with e.g. " snow" without looking up the numerical value of it.

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,45 @@
...
+    if let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
+        use gst::MessageView;
+        match msg.view() {
+            MessageView::Error(err) => {
+                eprintln!("Error received from element {}: {}",
+                         msg.get_src().get_path_string(), err.get_error());
+                eprintln!("Debugging information: {:?}", err.get_debug());
+            },
+            MessageView::Eos(..) => println!("End-Of-Stream reached."),
+            // We should not reach here because we only asked for ERRORs and EOS
```
sdroege: This comment is wrong: other than in the C code, we will reach here because the filtering of messages happens in our match statement and not before that already

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -0,0 +1,45 @@
...
+    source.set_property_from_str("pattern", "smpte");
+
+    // Start playing
+    let ret = pipeline.set_state(gst::State::Playing);
+    assert_ne!(ret, gst::StateChangeReturn::Failure,
+               "Unable to set the pipeline to the playing state.");
+
+    // Wait until error or EOS
+    let bus = pipeline.get_bus().unwrap();
+    if let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
```
sdroege: You want a loop around this to fetch all messages until EOS or Error, i.e. replace the `if let` with a `while let` and `break` on EOS and Error.

tutorials/src/bin/basic-tutorial-2.rs
```diff
@@ -29,16 +29,15 @@ fn main() {
...
     // Wait until error or EOS
     let bus = pipeline.get_bus().unwrap();
-    if let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
+    while let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
         use gst::MessageView;
         match msg.view() {
             MessageView::Error(err) => {
                 eprintln!("Error received from element {}: {}",
                          msg.get_src().get_path_string(), err.get_error());
                 eprintln!("Debugging information: {:?}", err.get_debug());
```
sdroege: On Error you also want to break. I'll fix that up and merge it, thanks!

Jouan: Oh, I completely missed that obvious part! 
Thank you for all your feedback, it's super helpful!

sdroege: I'm glad to hear that it was useful for you :) Thanks again for this PR, it's merged now

PR #35, opened 2017-09-14
Define Sync and Send for Query...

fengalin:
... if this is acceptable.


sdroege:
Thanks, that was just an oversight from me. Did the same for Message and Event now

PR #36, opened 2017-09-17
Set version (1.12) for StackTraceFlags

fengalin:
If understand correctly, `StackTraceFlags` was added to `gstinfo.h` after version 1.10.

See also https://github.com/sdroege/gstreamer-sys/pull/2


fengalin:
Note: this change makes my project generate successfully on AppVeyor (MSYS is using gstreamer-1.10.2).

PR #39, opened 2017-10-09
Add basic-tutorial-4 from the GStreamer tutorials

lovebug356:
https://cgit.freedesktop.org/gstreamer/gst-docs/tree/examples/tutorials/basic-tutorial-4.c

There are however some things that are not so nice (but not related to the example):

- ClockTime is not printed as nice as GST_TIME_FORMAT (add std::fmt::Debug trait on ClockTime?)
- seek_simple does not accept a ClockTime but a i64 for the new position

### DIFF
--- a/tutorials/src/bin/basic-tutorial-4.rs
+++ b/tutorials/src/bin/basic-tutorial-4.rs
@@ -1,0 +1,109 @@
+extern crate gstreamer as gst;
+use std::io;
+use std::io::Write;
+use gst::prelude::*;
+use gst::MessageView;
+
+fn main() {
+    // Initialize GStreamer
+    gst::init().unwrap();
+
+    // Creat the playbin element
+    let playbin = gst::ElementFactory::make("playbin", "playbin")
+        .expect("Failed to create playbin element");
+
+    // Set the URI to play
+    let uri = "https://www.freedesktop.org/software/gstreamer-sdk/data/media/sintel_trailer-480p.webm";
+    playbin.set_property("uri", &uri)
+        .expect("Can't set uri property on playbin");
+
+    // Start playing
+    let ret = playbin.set_state(gst::State::Playing);
+    assert_ne!(ret, gst::StateChangeReturn::Failure);
+
+    // Listen to the bus
+    let bus = playbin.get_bus().unwrap();
+    let mut is_playing = false;
+    let mut seek_enabled = false;
+    let mut seek_done = false;
+
+    loop {
+        let msg = bus.timed_pop(100 * gst::MSECOND);
+
+        match msg {
+            Some(msg) => {
+                match msg.view() {
+                    MessageView::Eos(..) => {
+                        println!("End-Of-Stream reached.");
+                        break
+                    },
+                    MessageView::Error(err) => {
+                        println!(
+                            "Error from {}: {} ({:?})",
+                            msg.get_src().get_path_string(),
+                            err.get_error(),
+                            err.get_debug()
+                        );
+                        break
+                    }
+                    MessageView::StateChanged(state) => {
+                        if msg.get_src() == playbin {
+                            let new_state = state.get_current();
+                            let old_state = state.get_old();
+
+                            println!("Pipeline state changed from {:?} to {:?}",
+                                     old_state, new_state);
+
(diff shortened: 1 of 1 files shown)


sdroege:
> ClockTime is not printed as nice as GST_TIME_FORMAT (add std::fmt::Debug trait on ClockTime?)

See https://github.com/gtk-rs/gir/issues/476 :) I'd like to implement `ClockTime` as `struct ClockTime(pub Option<u64>)` to properly handle `gst::CLOCK_TIME_NONE`. And then we can also implement `Debug` and `Display` nicely on it (we can't currently because of the orphan impl rules).

> seek_simple does not accept a ClockTime but a i64 for the new position

Just like in C, same for position and duration query. I think this is very unfortunate but there is not that much we can do about it.

tutorials/src/bin/basic-tutorial-4.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    // Start playing
+    let ret = playbin.set_state(gst::State::Playing);
+    assert_ne!(ret, gst::StateChangeReturn::Failure);
+
+    // Listen to the bus
+    let bus = playbin.get_bus().unwrap();
+    let mut is_playing = false;
+    let mut seek_enabled = false;
+    let mut seek_done = false;
```
sdroege: Maybe add some comments what these 3 variables *mean*.

There also must be a nicer way for doing this than having 3 mutable boolean flags, but I also currently don't see an easy way for that. By using futures this might be nicer, you could wait for playing, then do the query and schedule 10s in the future the seek.

But as this should be similar to the tutorial in C (so that the text can be reused), it should be fine. If you want you can submit a second version that is nicer though :)

lovebug356:
Added some comments (from the original C-files) and extracting the handle_message function to be more close to the original C-source files.

tutorials/src/bin/basic-tutorial-4.rs
```diff
@@ -0,0 +1,138 @@
+extern crate gstreamer as gst;
+use std::io;
+use std::io::Write;
+use gst::prelude::*;
+use gst::MessageView;
+
+struct CustomData {
```
sdroege: I don't like the name, but that's how it is in C so whatever... :)

tutorials/src/bin/basic-tutorial-4.rs
```diff
@@ -0,0 +1,138 @@
...
+    playbin: gst::Element, // Our one and only element
+    playing: bool, // Are we in the PLAYING state?
+    terminate: bool, // Should we terminate execution?
+    seek_enabled: bool, // Is seeking enabled for this media?
+    seek_done: bool, // Have we performed the seek already?
```
sdroege: These bool flags actually encode a state machine, right? So ideally that could be an enum for the current state. But let's keep it like it is in C

tutorials/src/bin/basic-tutorial-4.rs
```diff
@@ -0,0 +1,138 @@
...
+                    print!("\rPosition {} / {}", position, custom_data.duration);
+                    io::stdout().flush().unwrap();
+
+                    if custom_data.seek_enabled
+                        && !custom_data.seek_done
+                        && position > 10 * gst::SECOND {
+                        println!("\nReached 10s, performing seek...");
+                        custom_data.playbin.seek_simple(gst::Format::Time,
+                                            gst::SeekFlags::FLUSH | gst::SeekFlags::KEY_UNIT,
+                                            (30 * gst::SECOND) as i64)
```
sdroege: Maybe we could change this at a later time to an
```rust
enum FormattedPosition {
    Time(ClockTime),
    Percentage(u32),
    [...]
    Other(gst::Format, i64),
}
```
And use that for these generic functions that handle position and format together (and e.g. also the duration query).

tutorials/src/bin/basic-tutorial-4.rs
sdroege: @lovebug356 Can you also add a PR for tutorial 3 that adds this struct for more similarity with the C code?

lovebug356: Ok, I will have a look at it.

PR #40, opened 2017-10-10
gstreamer-base: use g_boxed_copy/free for GstFlowCombiner

fengalin:
This avoids depending on gst_flow_combiner_ref/unref which were introduced in gstreamer-1.12.1.


gstreamer-base/src/flow_combiner.rs
```diff
@@ -11,11 +16,15 @@ use std::mem;
 use std::ptr;
 
 glib_wrapper! {
-    pub struct FlowCombiner(Shared<ffi::GstFlowCombiner>);
+    pub struct FlowCombiner(Boxed<ffi::GstFlowCombiner>);
```
sdroege: It still has to be Shared, the boxed copy/free functions wrap around reference counting (unfortunately).

fengalin: I must admit I have difficulties being sure of the implications since it involves concepts from multiple frameworks that I'm still discovering. I gave a look at these types and it seemed to me that `g_boxed_copy` will actually perform a copy of the boxed object. Is it what we intend to do for `GstFlowCombiner`s? It seems to contradict the `SharedMemoryManager` trait.

Is this what this should look like:
``` rust
glib_wrapper! {
    pub struct FlowCombiner(Shared<ffi::GstFlowCombiner>);

    match fn {
        ref => |ptr| {
            gobject_ffi::g_boxed_copy(ffi::gst_flow_combiner_get_type(), ptr as *mut _) as *mut ffi::GstFlowCombiner
        },
        unref => |ptr| {
            gobject_ffi::g_boxed_free(ffi::gst_flow_combiner_get_type(), ptr as *mut _)
        },
        get_type => || ffi::gst_flow_combiner_get_type(),
    }
}
```

fengalin: Oh! I think I missed the last part of your comment (again...), the one about reference counting for Boxed types. I'll push a squashed version, hopefully correct this time.

Gir_GstBase.toml
```diff
@@ -67,7 +67,7 @@ concurrency = "none"
 
 [[object]]
 name = "GstBase.FlowCombiner"
-status = "generate"
+status = "manual"
```
sdroege: Add a comment why it is manual here

PR #42, opened 2017-10-12
Add basic-tutorial-7.rs

thiblahute:


tutorials/src/bin/basic-tutorial-7.rs
```diff
@@ -0,0 +1,66 @@
...
+    let video_queue = gst::ElementFactory::make("queue", "video_queue").unwrap();
+    let visual = gst::ElementFactory::make("wavescope", "visual").unwrap();
+    let video_convert = gst::ElementFactory::make("videoconvert", "video_convert").unwrap();
+    let video_sink = gst::ElementFactory::make ("autovideosink", "video_sink").unwrap();
+
+    let pipeline = gst::Pipeline::new("test-pipeline");
+
+    audio_source.set_property("freq", &215.0).unwrap();
+    visual.set_property_from_str("shader", "0");
+    visual.set_property_from_str("style", "1");
```
sdroege: I'd prefer using `set_property()` always. Similar to here https://github.com/sdroege/gstreamer-rs/blob/b773c0b46a75ecccc21841abf317d94ea2a116e4/examples/src/bin/playbin.rs#L24 you can also do that with glib::EnumClass

tutorials/src/bin/basic-tutorial-7.rs
```diff
@@ -0,0 +1,66 @@
...
+    println!("Obtained request pad {} for audio branch", tee_audio_pad.get_name());
+    let queue_audio_pad = audio_queue.get_static_pad("sink").unwrap();
+    tee_audio_pad.link(&queue_audio_pad);
+
+    let tee_video_pad = tee.get_request_pad("src_%u").unwrap();
+    println!("Obtained request pad {} for video branch", tee_video_pad.get_name());
+    let queue_video_pad = video_queue.get_static_pad("sink").unwrap();
+    tee_video_pad.link(&queue_video_pad);
+
+    pipeline.set_state(gst::State::Playing);
```
sdroege: Don't we want to check if it succeeds? :)

tutorials/src/bin/basic-tutorial-7.rs
thiblahute: Well, this is much more code for the same thing, I would rather adviertise using `set_property_from_str` tbh.

tutorials/src/bin/basic-tutorial-7.rs
thiblahute: We want a `Result` ;) -- But yes, of course.

sdroege: See my comment about that in #gstreamer earlier

tutorials/src/bin/basic-tutorial-7.rs
sdroege: Via GEnumClass you can get proper error handling, this here more or less silently fails. I'm fine with keeping it, but then please use the proper name ("none" and "lines") instead of the integer values :)

thiblahute: Well yeah, but has you told me yesterday, this never fails unless you fail at writing :-)

Issue #45, opened 2017-10-22
Clean the tutorials (basic-tutorial-6)


luisbg:
Make the tutorial code more readable and idiomatic. A few small issues in the basic-tutorial-6.

PR #46, opened 2017-10-22
Clean tutorial 6

luisbg:
Fixes for Issue #45

### DIFF
--- a/tutorials/src/bin/basic-tutorial-6.rs
+++ b/tutorials/src/bin/basic-tutorial-6.rs
@@ -3,7 +3,7 @@ use gst::prelude::*;
 use gst::MessageView;
 
 
-fn print_caps(caps: gst::Caps, prefix: &str) {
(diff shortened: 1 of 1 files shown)


tutorials/src/bin/basic-tutorial-6.rs
```diff
@@ -122,39 +122,36 @@ fn main() {
     loop {
         let msg = bus.timed_pop(gst::CLOCK_TIME_NONE);
 
-        match msg {
-            Some(msg) => {
-                match msg.view() {
-                    MessageView::Error(err) => {
+        if let Some(msg) = msg {
```
sdroege: You can even get rid of the loop here.
```rust
while let Some(msg) = bus.timed_pop(...) {
  [...]
}
```

luisbg: True!

Didn't know about while let. Cool.

Will update.

PR #51, opened 2017-11-04
video: Add wrappers for GstVideoDecoder and its dependencies

thiblahute:
This is a plain simple wrapping of the C API.


sdroege:
Seems like you forgot to add a file, travis complains :)

thiblahute:
> Seems like you forgot to add a file, travis complains :)

Oops, should be better now :-)

gstreamer-video/src/video_info.rs
```diff
@@ -166,6 +166,14 @@ impl fmt::Display for ::VideoColorimetry {
 
 pub struct VideoInfo(pub(crate) ffi::GstVideoInfo);
 
+impl fmt::Debug for VideoInfo {
```
sdroege: This should probably be a separate commit

gstreamer-video/src/video_codec_state.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub struct VideoCodecState(Shared<ffi::GstVideoCodecState>);
+
+    match fn {
+        ref => |ptr| ffi::gst_video_codec_state_ref(ptr),
+        unref => |ptr| ffi::gst_video_codec_state_unref(ptr),
+        get_type => || ffi::gst_video_codec_state_get_type(),
+    }
+}
+
+impl VideoCodecState {
```
sdroege: This API is not enough, subclasses need to be able to modify the fields of the codec state

gstreamer-video/src/video_codec_frame.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub struct VideoCodecFrame(Shared<ffi::GstVideoCodecFrame>);
+
+    match fn {
+        ref => |ptr| ffi::gst_video_codec_frame_ref(ptr),
+        unref => |ptr| ffi::gst_video_codec_frame_unref(ptr),
+        get_type => || ffi::gst_video_codec_frame_get_type(),
+    }
+}
+
+impl VideoCodecFrame {
```
sdroege: Subclasses need to be able to modify the fields of this, which is unfortunately not thread-safe (and thus `Send` and `Sync` and not exactly correct)

gstreamer-video/src/auto/video_decoder.rs
```diff
@@ -0,0 +1,295 @@
...
+    match fn {
+        get_type => || ffi::gst_video_decoder_get_type(),
+    }
+}
+
+unsafe impl Send for VideoDecoder {}
+unsafe impl Sync for VideoDecoder {}
+
+pub trait VideoDecoderExt {
+    fn add_to_frame(&self, n_bytes: i32);
```
sdroege: This, and various other functions must only be called while having taken the decoder stream lock

gstreamer-video/src/auto/video_decoder.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    fn get_packetized(&self) -> bool;
+
+    fn get_pending_frame_size(&self) -> usize;
+
+    fn get_qos_proportion(&self) -> f64;
+
+    fn have_frame(&self) -> gst::FlowReturn;
+
+    //fn merge_tags<'a, P: Into<Option<&'a /*Ignored*/gst::TagList>>>(&self, tags: P, mode: /*Ignored*/gst::TagMergeMode);
```
sdroege: You get these generated if you add taglist and tag merge mode as manual into the toml file. See how it's done for taglist (or caps, structure, etc) in the other toml files.

gstreamer-video/src/auto/video_decoder.rs
thiblahute: So I take the lock myself in the imple, or how would you do that?

gstreamer-video/src/video_info.rs
thiblahute: Done.

gstreamer-video/src/video_codec_state.rs
thiblahute: Done.

gstreamer-video/src/video_codec_frame.rs
thiblahute: Yeah, should we figure out a proper way to do it right now? Do you have ideas to investigate?

gstreamer-video/src/auto/video_decoder.rs
thiblahute: Done.

gstreamer-video/src/auto/video_decoder.rs
sdroege: I would have all methods that require the stream lock to be taken to do that inside the implementation. It's a recursive mutex, so doing it one time too often is not a problem apart from performance, and it would make this safe at least.

At a later time we could consider exposing unsafe functions that don't do that, for being able to implement a more performant solution. And also unsafe functions for unlocking the mutex (which is required if you want to block inside handle_frame()).

gstreamer-video/src/video_codec_frame.rs
sdroege: For now I would do a manual implementation for the getters/setters.

Not sure what to do about the thread-safety issue, as nothing is going to prevent you from getting a new reference at any time from another thread. Maybe another layer of wrapping would be needed, so that accessing any content of the frame would take the stream lock too.

gstreamer-video/src/video_codec_state.rs
sdroege: Same thing as for the VideoCodecFrame. Some kind of locking would be necessary here, and I think the same approach would work for now.

gstreamer-video/src/video_info.rs
```diff
@@ -166,6 +166,14 @@ impl fmt::Display for ::VideoColorimetry {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?} - ({:?}x{:?} - {:?}bytes)", self.format(),
+            self.width(), self.height(), self.size())
+    }
+}
```
sdroege: Once I regenerate gstreamer-sys soonish, you should be able to simply #[derive(Debug)] (ffi::GstVideoInfo would implement Debug already).

gstreamer-video/src/video_codec_frame.rs
sdroege: Could also everywhere where you can get a frame, instead return a struct that has a reference to the actual frame and the decoder, and on lock() locks the stream lock and returns a mutex guard that on Deref returns a frame reference

Same for codec state

sdroege:
Any progress here?

thiblahute:
> Any progress here?

Not yet, I will try to work on it again soon!

sdroege:
This is now tracked at https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/merge_requests/51

PR #53, opened 2017-11-10
playbin: More robust cli arg handling

zeenix:


examples/src/bin/playbin.rs
```diff
@@ -9,10 +9,11 @@ fn main() {
     gst::init().unwrap();
 
     let args: Vec<_> = env::args().collect();
-    let uri: &str = if args.len() == 2 {
+    let uri: &str = if args.len() >= 2 {
```
sdroege: If you provide more than 2, it will ignore all others. Shouldn't it rather fail and say that only one argument should be provided.

zeenix: hmm.. true

zeenix: fixed

sdroege:
Ok, looks good now. But can you do the same for the other examples too then?

zeenix:
yup, done.

PR #62, opened 2017-11-12
Set up a runloop on macOS

### LINKED ISSUE #58: Tutorials don't work on macOS due to not running a NSRunLoop on the main thread, as required for OpenGL on macOS
```sh
gstreamer-rs/tutorials$ cargo run --bin basic-tutorial-1
```

> couldn't load TLS file database: Failed to open file '/Library/Frameworks/GStreamer.framework/Versions/1.0/etc/ssl/certs/ca-certificates.crt': No such file or directory

I've installed gstreamer-1.0-devel-1.12.3-x86_64.pkg, gstreamer-1.0-1.12.3-x86_64.pkg, and the dmg with a ton of installer packages (base-crypto-1.12.3-x86_64.pkg, base-system-1.0-1.12.3-x86_64.pkg, gstreamer-1.0-capture-1.12.3-x86_64.pkg, gstreamer-1.0-codecs-1.12.3-x86_64.pkg, gstreamer-1.0-codecs-gpl-1.12.3-x86_64.pkg, gstreamer-1.0-codecs-restricted-1.12.3-x86_64.pkg, gstreamer-1.0-core-1.12.3-x86_64.pkg, gstreamer-1.0-devtools-1.12.3-x86_64.pkg, gstreamer-1.0-dvd-1.12.3-x86_64.pkg, gstreamer-1.0-editing-1.12.3-x86_64.pkg, gstreamer-1.0-effects-1.12.3-x86_64.pkg, gstreamer-1.0-encoding-1.12.3-x86_64.pkg, gstreamer-1.0-libav-1.12.3-x86_64.pkg, gstreamer-1.0-net-1.12.3-x86_64.pkg, gstreamer-1.0-net-restricted-1.12.3-x86_64.pkg, gstreamer-1.0-playback-1.12.3-x86_64.pkg, gstreamer-1.0-system-1.12.3-x86_64.pkg, gstreamer-1.0-visualizers-1.12.3-x86_64.pkg, osx-framework-1.12.3-x86_64.pkg).

and yet, the directory "/Library/Frameworks/GStreamer.framework/Versions/1.0/etc/ssl/" is missing (there's only "fonts" in "etc").

pornel:
Fixes #58

It does make tutorial code a little bit uglier, but I don't know how to otherwise make it optional to run the code on a secondary thread.

### DIFF
--- a/tutorials/src/bin/basic-tutorial-4.rs
+++ b/tutorials/src/bin/basic-tutorial-4.rs
@@ -13,7 +13,11 @@ struct CustomData {
     duration: gst::ClockTime, // How long does this media last, in nanoseconds
 }
 
-fn main() {
+#[path="../tutorials-common.rs"]
+mod tutorials_common;
+
+fn tutorial_main() {
+
     // Initialize GStreamer
     gst::init().unwrap();
 
@@ -146,4 +150,10 @@ fn handle_message(custom_data: &mut CustomData, msg: &gst::GstRc<gst::MessageRef
         },
         _ => (),
     }
+}
+
+fn main() {
+    // tutorials_common::run is only required to set up the application environment on macOS
+    // (but not necessary in normal Cocoa applications where this is set up automatically)
+    tutorials_common::run(tutorial_main);
 }
--- a/tutorials/src/bin/basic-tutorial-1.rs
+++ b/tutorials/src/bin/basic-tutorial-1.rs
@@ -1,7 +1,11 @@
 extern crate gstreamer as gst;
 use gst::prelude::*;
 
-fn main() {
+#[path="../tutorials-common.rs"]
+mod tutorials_common;
+
+fn tutorial_main() {
+
     // Initialize GStreamer
     gst::init().unwrap();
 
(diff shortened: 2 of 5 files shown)


philn:
Thanks, I was about to start a similar patch!

sdroege:
> It does make tutorial code a little bit uglier, but I don't know how to otherwise make it optional to run the code on a secondary thread.

Yeah, no way to not make it ugly unfortunately :) Thanks for the patch!

tutorials/src/bin/basic-tutorial-4.rs
```diff
@@ -13,7 +13,17 @@ struct CustomData {
     duration: gst::ClockTime, // How long does this media last, in nanoseconds
 }
 
+#[path="../tutorials-common.rs"]
+mod tutorials_common;
+
 fn main() {
+    // tutorials_common::run is only required to set up the application environment on macOS
+    // (but not necessary in normal Cocoa applications where this is set up automatically)
+    tutorials_common::run(tutorial_main);
```
sdroege: I would do this slightly different (also don't copy the explanation into every tutorial, make it just a line :) ). I would make it some kind of `tutorials_common::init()` that is the first thing in main. Doing it with a closure that calls the actual main is nice, but it adds more compliciation to the tutorial IMHO

The same change is also needed for everything in the examples subdirectory, for everything that is showing video. Also tutorial 7 needs this too.

sdroege: Actually ignore this comment. You're moving the actual tutorial code in a different thread. Because the NSRunLoop has to run on the main thread.

Then we only need to add this also to tutorial 7 and the examples.

pornel:
I've moved main below, so at least the code reads nicer.

sdroege:
Thanks, I'll merge that in a bit :)

PR #65, opened 2017-12-10
gstreamer-player: regen for overridden signals

philn:
In the previous generated player bindings the position-updated was Send+Sync but
only Send is actually needed. In addition to this change the position-updated
and seek-done signals now invoke their handler with a gst::ClockTime value
instead of u64.

### DIFF
--- a/Gir_GstPlayer.toml
+++ b/Gir_GstPlayer.toml
@@ -78,6 +78,8 @@ trait = false
     [[object.signal]]
     name = "duration-changed"
     concurrency = "send"
+    # Pass ClockTime instead of u64
+    ignore = true
 
     [[object.signal]]
     name = "end-of-stream"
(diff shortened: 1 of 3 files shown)


gstreamer-player/src/auto/player.rs
```diff
@@ -282,8 +282,7 @@ impl Player {
 
     pub fn get_property_suburi(&self) -> Option<String> {
         unsafe {
-            let mut value = Value::uninitialized();
-            gobject_ffi::g_value_init(value.to_glib_none_mut().0, <String as StaticType>::static_type().to_glib());
+            let mut value = Value::from_type(<String as StaticType>::static_type());
```
sdroege: This requires GIT master for glib-rs, I'd prefer not to depend on that as I want to make a 0.10 release soonish without waiting for glib-rs to be released first. Can you use the version of gir that was used last time, d50d839.

Gir_GstPlayer.toml
```diff
@@ -96,12 +96,16 @@ trait = false
...
+    name = "position-updated"
     concurrency = "send"
+    # Pass ClockTime instead of u64
+    ignore = true
 
     [[object.signal]]
     name = "seek-done"
     concurrency = "send"
+    # Pass ClockTime instead of u64
+    ignore = true
```
sdroege: Also duration-changed

philn: Done :)

PR #66, opened 2017-12-22
Pad::remove_probe: don't capture PadProbeId

fengalin:
Use case: add a pad_probe while building the pipeline, then remove it in the bus watch's closure.

### DIFF
--- a/gstreamer/src/pad.rs
+++ b/gstreamer/src/pad.rs
@@ -105,7 +105,7 @@ pub trait PadExtManual {
     fn add_probe<F>(&self, mask: PadProbeType, func: F) -> PadProbeId
     where
         F: Fn(&Pad, &mut PadProbeInfo) -> PadProbeReturn + Send + Sync + 'static;
-    fn remove_probe(&self, id: PadProbeId);
+    fn remove_probe(&self, id: &PadProbeId);
 
     fn chain(&self, buffer: Buffer) -> FlowReturn;
     fn push(&self, buffer: Buffer) -> FlowReturn;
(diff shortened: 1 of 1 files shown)


sdroege:
It is intentionally done the way it is so that you can't possibly try to remove the same PadProbeId twice (where the second one could be another pad probe that someone else added in the meantime).

Can you show me some code that this currently makes awkward?

sdroege:
It's basically the same as with glib::SourceId

fengalin:
Ex.:

``` rust
    [...]
    let mut init_done = false;
    let toc_probe_id =
        audio_enc_sink_pad.add_probe(gst::PadProbeType::EVENT_DOWNSTREAM, |_pad, probe_info| {
            if let Some(ref data) = probe_info.data {
                match data {
                    &gst::PadProbeData::Event(ref event) => match event.view() {
                        gst::EventView::Toc(ref _toc) => return gst::PadProbeReturn::Drop,
                        _ => (),
                    },
                    _ => (),
                }
            }
            gst::PadProbeReturn::Ok
        });
    [...]

    self.pipeline.get_bus().unwrap().add_watch(move |_, msg| {
        match msg.view() {
            [...]
            gst::MessageView::AsyncDone(_) => {
                if !init_done {
                    init_done = true;
                     // Stop dropping TOCs
                    audio_enc_sink_pad.remove_probe(toc_probe_id);
            [...]
```

Compiler output:
```
error[E0507]: cannot move out of captured outer variable in an `FnMut` closure
   --> src/media/splitter_context.rs:215:57
    |
165 |         let toc_probe_id =
    |             ------------ captured outer variable
...
215 |                         audio_enc_sink_pad.remove_probe(toc_probe_id);
    |                                                         ^^^^^^^^^^^^ cannot move out of captured outer variable in an `FnMut` closure
```

sdroege:
That code would potentially remove the pad probe too many times, so the compiler complaining here is actually a good thing.

What I would suggest here is to store the probe ID in an `Option` and then doing
```rust
if let Some(id) = toc_probe_id.take() {
    pad.remove_probe(id);
}
```

fengalin:
Yes, I just figured that out... Sorry.

sdroege:
No problem :)

PR #69, opened 2017-12-26
gstreamer-video: VideoRectangle bindings

philn:
I needed this to center the rectangle passed to VideoOverlay::set_render_rect()

### DIFF
--- a/gstreamer-video/src/video_rectangle.rs
+++ b/gstreamer-video/src/video_rectangle.rs
@@ -1,0 +1,35 @@
+// Copyright (C) 2017 Philippe Normand <philn@igalia.com~
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use ffi;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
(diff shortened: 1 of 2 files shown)


gstreamer-video/src/video_rectangle.rs
```diff
@@ -0,0 +1,35 @@
...
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use ffi;
+
+#[derive(Debug)]
```
sdroege: This probably should get a few more traits derived. PartialEq, Eq for example. And you might want to implement the GLib conversion traits

sdroege: Actually the conversion traits wouldn't make much sense, considering how it is used. That would only work if you made it a wrapper around the FFI type.

philn: Ok, PR updated :)

sdroege:
Please run other manual code through rustfmt (rustfmt-nightly) in the future :) I did this now after merging because I forgot

philn:
Ah yes, sorry! It would be nice to have a pre-commit hook for this. I found this one for instance, https://github.com/google/tarpc/blob/master/hooks/pre-commit

sdroege:
Something running on the PRs would be nicer, but doing that on travis would take quite some time as you would have to build rustfmt. I saw some travis configurations for that though. Will have to think about that at some point :)

PR #71, opened 2018-01-12
Pad: allow pad probes to capture their environment as mutable

fengalin:
Example use case:

``` rust
    let mut is_done = AtomicBool::new(false);
    sink_pad.add_probe(gst::PadProbeType::BUFFER, move |_pad, probe_info| {
         let mut is_cndt_lck = is_cndt.get_mut();
         if !*is_done_lck {
             // Do stuff
             *is_done_lck = true;
         } else {
             // Do other stuff
         }
         ...
    }
```

I think this is valid because the function for `add_probe` is constrained to be `Send + Sync`.
If we only allow for `Fn` functions, the inner condition must be a clone to an `Arc<Mutex<bool>>`.

### DIFF
--- a/gstreamer/src/pad.rs
+++ b/gstreamer/src/pad.rs
@@ -104,7 +104,7 @@ impl Drop for StreamLock {
 pub trait PadExtManual {
     fn add_probe<F>(&self, mask: PadProbeType, func: F) -> PadProbeId
     where
-        F: Fn(&Pad, &mut PadProbeInfo) -> PadProbeReturn + Send + Sync + 'static;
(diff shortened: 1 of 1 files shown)


sdroege:
You forgot to update https://github.com/sdroege/gstreamer-rs/blob/4297f8f5c58cb6dfd8864a069482ddf777a3cd6c/gstreamer/src/pad.rs#L855 and then you would've seen that it's not safe :) You would create a mutable reference to the closure here, and Rust does not allow two mutable references to the same thing (which would be there if the pad probe was called twice at the same time).

fengalin:
Yeah, you're right! :)

PR #72, opened 2018-01-16
Document dependency for libgstreamer-plugins-bad1.0-dev

charlie-ht:
I saw this after doing a fresh checkout on Ubuntu 17.10,

```
error: failed to run custom build command for `gstreamer-player-sys v0.5.0 (https://github.com/sdroege/gstreamer-sys#d68a12a4)`
process didn't exit successfully: `/home/charlie/git/gstreamer-rs/target/debug/build/gstreamer-player-sys-8838fb523b245540/build-script-build` (exit code: 1)
--- stderr
`"pkg-config" "--libs" "--cflags" "gstreamer-player-1.0 >= 1.12"` did not exit successfully: exit code: 1
--- stderr
Package gstreamer-player-1.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gstreamer-player-1.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'gstreamer-player-1.0' found


warning: build failed, waiting for other jobs to finish...
error: build failed
```

Add this dependency fixed the issue here. I assume the same will be true on Debian, but I haven't tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,22 @@ On Debian/Ubuntu they can be installed with
 
 ```
 $ apt-get install libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev \
-      libgstreamer-plugins-bad1.0-dev \
       gstreamer1.0-plugins-base gstreamer1.0-plugins-good \
       gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly \
       gstreamer1.0-libav
+```
+
+The minimum required version of the above libraries is >= 1.8. If you
+build the gstreamer-player sub-crate, or any of the examples that
+depend on gstreamer-player, you must ensure that in addition to the
+above packages, `libgstreamer-plugins-bad1.0-dev` is installed and
+that the version is >= 1.12. See the `Cargo.toml` files for the full
+details,
+
+```
+# Only if you wish to install gstreamer-player, make sure the version
+# of this package is >= 1.12.
+$ apt-get install libgstreamer-plugins-bad1.0-dev
 ```
 
 Package names on other distributions should be similar.
@@ -69,6 +81,10 @@ provided by the GStreamer project.
 $ brew install gstreamer gst-plugins-base gst-plugins-good \
       gst-plugins-bad gst-plugins-ugly gst-libav
 ```
+
+If you wish to install the gstreamer-player sub-crate, make sure the
+version of these libraries is >= 1.12. Otherwise, a version >= 1.8 is
+sufficient.
 
 #### GStreamer Binaries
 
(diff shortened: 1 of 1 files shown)


charlie-ht:
Oof, I don't have any good ideas either :)  I don't suppose the requirement could be >= 1.12 for the whole library?

It does seem wrong to me to omit a required dependency in general.

sdroege:
> Oof, I don't have any good ideas either :) I don't suppose the requirement could be >= 1.12 for the whole library?

That would limit the usefulness a lot. Many distros still have 1.10 or 1.8.

> It does seem wrong to me to omit a required dependency in general.

Maybe add another command-line to install only -bad and write above that this is only needed for the gstreamer-player crate (and that 1.12 is required for that). Please also add the equivalent for the other platforms (msys2, homebrew).

charlie-ht:
If they only have 1.8 or 1.10, then they can't easily install the 1.12 version required for the player example, so adding the obvious apt-get incantation is only going to frustrate those who couldn't already figure out what package to install from the first error message.

I also wasn't aware that Homebrew separated out -bad into -bad (lib files) and -dev (lib files + headers), I thought the -bad package installed them both, but I've not used Homebrew or the pacman packages to know (pacman looked similar in it's behaviour to homebrew)

I don't understand Rust or Cargo well enough at this point to confidently document what dependencies should be installed for what purposes, so I'll slowly back away for now. :)

sdroege:
I don't know either, I'm not good with docs :)

I would suggest to add to the README.md that 1.8 is needed and you install it like written so far, and that in addition for gstreamer-player libgstreamer1.0-plugins-bad-dev 1.12 is needed. And various of the examples/tutorials have additional dependencies (including e.g. GTK) but hidden behind a feature-gate, and that the Cargo.toml of the examples/tutorials should be checked for details.

What do you think? What that have solved your initial confusion?

charlie-ht:
> What do you think? What that have solved your initial confusion?

To be honest, I wasn't initially that confused by the error, knew it was a missing package. I thought it would be helpful to add a missing package. If I'd have known about the library differences before starting out, I would have perhaps not proposed this docs change :)

I still think this is going to be a headache for those stuck on 1.8 and wishing to build gstreamer-player, but at least they've been amply warned of an incoming headache now.

sdroege:
Well, there simply is no GstPlayer in 1.8 :) Thanks for the update, I'll review in a bit

charlie-ht:
Ahh oops, the distinction makes a lot more sense now, I initially thought it was some API incompatibility :) Thank you for the insight.

sdroege:
1.10 has GstPlayer but there are API differences

PR #73, opened 2018-01-19
Rtpfec example

MathieuDuponchelle:
This depends on https://bugzilla.gnome.org/show_bug.cgi?id=792696

### DIFF
--- a/examples/src/bin/rtpfecclient.rs
+++ b/examples/src/bin/rtpfecclient.rs
@@ -1,0 +1,290 @@
+#[macro_use]
+extern crate gstreamer as gst;
+use gst::prelude::*;
+
+extern crate glib;
+
+use std::env;
+use std::error::Error as StdError;
+use std::sync::{Arc, Mutex};
+
+#[path = "../examples-common.rs"]
+mod examples_common;
+
+extern crate failure;
+use failure::Error;
+
+#[macro_use]
+extern crate failure_derive;
+
+#[derive(Debug, Fail)]
+#[fail(display = "Missing element {}", _0)]
+struct MissingElement(&'static str);
+
+#[derive(Debug, Fail)]
+#[fail(display = "No such pad {} in {}", _0, _1)]
+struct NoSuchPad(&'static str, String);
+
+#[derive(Debug, Fail)]
+#[fail(display = "Unknown payload type {}", _0)]
+struct UnknownPT(u32);
+
+#[derive(Debug, Fail)]
+#[fail(display = "Usage: {} (play | record) DROP_PROBABILITY", _0)]
+struct UsageError(String);
+
+#[derive(Debug, Fail)]
+#[fail(display = "Received error from {}: {} (debug: {:?})", src, error, debug)]
+struct ErrorMessage {
+    src: String,
+    error: String,
+    debug: Option<String>,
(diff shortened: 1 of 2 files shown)


sdroege:
The 1.10 dependency should be conditional, like in the decodebin example. Many distros still have 1.8 unfortunately and you even update the requirement for all examples :)

MathieuDuponchelle:
> The 1.10 dependency should be conditional, like in the decodebin example. Many distros still have 1.8 unfortunately and you even update the requirement for all examples :)

Not sure why I needed it tbh, just removed it and things still build fine, updated the PR :)

examples/src/bin/rtpfecclient.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+fn make_fec_decoder (rtpbin: &gst::Element, sess_id: u32) -> Result<gst::Element, Error> {
+    let fecdec = make_element("rtpulpfecdec", None)?;
+    let internal_storage = rtpbin.emit("get-internal-storage", &[&sess_id.to_value()]).unwrap().unwrap();
+
+    fecdec.set_property("storage", &internal_storage.to_value())?;
+    fecdec.set_property("pt", &100u32.to_value())?;
+
+    Ok (fecdec)
```
sdroege: Please run this through rustfmt-nightly :)

examples/src/bin/rtpfecclient.rs
```diff
@@ -0,0 +1,290 @@
...
+            eprintln!("Recording to out.mkv");
+        },
+        _ => return Err(Error::from(UsageError(args[0].clone())))
+    }
+
+    src.link(&netsim)?;
+
+    rtpbin.connect("new-storage", false, |values| {
+        let storage = values[1].get::<gst::Element>().expect("Invalid argument");
+        storage.set_property("size-time", &250000000u64.to_value()).unwrap();
```
sdroege: This is looking for some underscores: `250_000_000`

examples/src/bin/rtpfecclient.rs
```diff
@@ -0,0 +1,290 @@
...
+    assert_ne!(ret, gst::StateChangeReturn::Failure);
+
+    while let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
+        use gst::MessageView;
+
+        match msg.view() {
+            MessageView::Eos(..) => break,
+            MessageView::Error(err) => {
+                {
+                    match err.get_details() {
```
sdroege: Hm this should be 1.10 API?

examples/src/bin/rtpfecserver.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    while let Some(msg) = bus.timed_pop(gst::CLOCK_TIME_NONE) {
+        use gst::MessageView;
+
+        match msg.view() {
+            MessageView::Eos(..) => break,
+            MessageView::Error(err) => {
+                {
+                    match err.get_details() {
+                        Some(details) if details.get_name() == "error-details" => details
```
sdroege: In both applications there is nothing that would ever have Some details. You don't post any error message with details anywhere

sdroege:
Doing the same example without FEC might also be nice for a as-simple-as-possible rtpbin example

PR #75, opened 2018-01-25
Fix building argument-less messages

fengalin:
Building an argument-less message such as `eos` yields an assertion failure due to the inner `structure` being `null`.

Events are not affected as the `build` method uses `gst_event_writable_structure` and this function "will never return NULL".

### DIFF
--- a/gstreamer/src/event.rs
+++ b/gstreamer/src/event.rs
@@ -895,7 +895,7 @@ macro_rules! event_builder_generic_impl {
                     ffi::gst_event_set_running_time_offset(event, running_time_offset);
                 }
 
-                {
+                if !self.other_fields.is_empty() {
                     let s = StructureRef::from_glib_borrow_mut(
                         ffi::gst_event_writable_structure(event)
                     );
@@ -1666,4 +1666,50 @@ impl<'a> CustomBothOobBuilder<'a> {
 
         ev
     });
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_simple() {
+        ::init().unwrap();
+
+        // Event without arguments
+        let flush_start_evt = Event::new_flush_start().build();
+        match flush_start_evt.view() {
+            EventView::FlushStart(flush_start_evt) => {
+                assert!(flush_start_evt.0.get_structure().is_none());
+            },
+            _ => panic!("flush_start_evt.view() is not an EventView::FlushStart(_)"),
+        }
+
+        let flush_start_evt = Event::new_flush_start()
+            .other_fields(&[("extra-field", &true)])
+            .build();
+        match flush_start_evt.view() {
+            EventView::FlushStart(flush_start_evt) => {
+                assert!(flush_start_evt.0.get_structure().is_some());
+                if let Some(other_fields) = flush_start_evt.0.get_structure() {
+                    assert!(other_fields.has_field("extra-field"));
+                }
+            },
+            _ => panic!("flush_start_evt.view() is not an EventView::FlushStart(_)"),
+        }
+
+        // Event with arguments
+        let flush_stop_evt = Event::new_flush_stop(true)
+            .other_fields(&[("extra-field", &true)])
+            .build();
+        match flush_stop_evt.view() {
+            EventView::FlushStop(flush_stop_evt) => {
+                assert_eq!(flush_stop_evt.get_reset_time(), true);
+                assert!(flush_stop_evt.0.get_structure().is_some());
+                if let Some(other_fields) = flush_stop_evt.0.get_structure() {
+                    assert!(other_fields.has_field("extra-field"));
+                }
+            }
+            _ => panic!("flush_stop_evt.view() is not an EventView::FlushStop(_)"),
+        }
+    }
 }
--- a/gstreamer/src/message.rs
+++ b/gstreamer/src/message.rs
@@ -1165,16 +1165,15 @@ macro_rules! message_builder_generic_impl {
             }
         }
 
-        // Warning: other_fields are ignored with argument-less messages
-        // until GStreamer 1.14 is released
-        pub fn other_fields(self, other_fields: &[(&'a str, &'a ToSendValue)]) -> Self {
+        // TODO: restore other_fields method and condition it to the "v1_14" feature
+        /*pub fn other_fields(self, other_fields: &[(&'a str, &'a ToSendValue)]) -> Self {
             Self {
                 other_fields: self.other_fields.iter().cloned()
                     .chain(other_fields.iter().cloned())
                     .collect(),
                 .. self
             }
-        }
+        }*/
 
         pub fn build(mut self) -> Message {
             assert_initialized_main_thread!();
@@ -2468,19 +2467,16 @@ mod tests {
             _ => panic!("eos_msg.view() is not a MessageView::Eos(_)"),
         }
 
-        // Note: can't define other_fields for argument-less messages before GStreamer 1.14
-
         // Message with arguments
-        let buffering_msg = Message::new_buffering(42)
-            .other_fields(&[("extra-field", &true)])
-            .build();
+        let buffering_msg = Message::new_buffering(42).build();
         match buffering_msg.view() {
             MessageView::Buffering(buffering_msg) => {
                 assert_eq!(buffering_msg.get_percent(), 42);
-                assert!(buffering_msg.0.get_structure().is_some());
-                assert!(buffering_msg.0.get_structure().unwrap().has_field("extra-field"));
             }
             _ => panic!("buffering_msg.view() is not a MessageView::Buffering(_)"),
         }
+
+        // TODO: add a tests for messages with and without arguments with an extra-field
+        // and condition it to the "v1_14" feature
     }
 }
--- a/gstreamer/src/message.rs
+++ b/gstreamer/src/message.rs
@@ -1165,6 +1165,8 @@ macro_rules! message_builder_generic_impl {
             }
         }
 
+        // Warning: other_fields are ignored with argument-less messages
+        // until GStreamer 1.14 is released
         pub fn other_fields(self, other_fields: &[(&'a str, &'a ToSendValue)]) -> Self {
             Self {
                 other_fields: self.other_fields.iter().cloned()
@@ -1183,13 +1185,19 @@ macro_rules! message_builder_generic_impl {
                     ffi::gst_message_set_seqnum(msg, seqnum.to_glib());
                 }
 
-                {
-                    let s = StructureRef::from_glib_borrow_mut(
-                        ffi::gst_message_get_structure(msg) as *mut _
-                    );
+                if !self.other_fields.is_empty() {
+                    // issue with argument-less messages. We need the function
+                    // ffi::gst_message_writable_structure to sort this out
+                    // and this function will be available in GStreamer 1.14
+                    // See https://github.com/sdroege/gstreamer-rs/pull/75
+                    // and https://bugzilla.gnome.org/show_bug.cgi?id=792928
+                    let structure = ffi::gst_message_get_structure(msg);
+                    if !structure.is_null() {
+                        let structure = StructureRef::from_glib_borrow_mut(structure as *mut _);
 
-                    for (k, v) in self.other_fields {
-                        s.set_value(k, v.to_send_value());
+                        for (k, v) in self.other_fields {
+                            structure.set_value(k, v.to_send_value());
+                        }
                     }
                 }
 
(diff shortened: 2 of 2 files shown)


gstreamer/src/event.rs
```diff
@@ -1667,3 +1667,27 @@ impl<'a> CustomBothOobBuilder<'a> {
...
+        let flush_start_evt = Event::new_flush_start().build();
+        match flush_start_evt.view() {
+            EventView::FlushStart(_) => (),
+            _ => panic!("flush_start_evt.view() is not an EventView::FlushStart(_)"),
+        }
+
+        // Event with arguments
+        let flush_stop_evt = Event::new_flush_stop(true).build();
+        match flush_stop_evt.view() {
+            EventView::FlushStop(flush_stop_evt) => assert!(flush_stop_evt.get_reset_time()),
```
sdroege: What exactly is this assertion doing?

gstreamer/src/message.rs
```diff
@@ -1184,12 +1184,13 @@ macro_rules! message_builder_generic_impl {
                 }
 
                 {
-                    let s = StructureRef::from_glib_borrow_mut(
-                        ffi::gst_message_get_structure(msg) as *mut _
-                    );
+                    let structure = ffi::gst_message_get_structure(msg);
+                    if !structure.is_null() {
+                        let structure = StructureRef::from_glib_borrow_mut(structure as *mut _);
```
sdroege: This means that extra fields are now silently ignored. It's the best we can do without breaking API now, but for the next API breakage the other fields API should just be removed for messages. Doing it properly requires additions to GStreamer (similar to what exists for events/queries).

I'll merge your patch to the 0.10 branch (once the question above is solved) and make a new release soon, and remove the API in master.

gstreamer/src/event.rs
fengalin: Checking that the `reset_time` is `true` as it was initialized when building the event. This is on par with `assert_eq!(buffering_msg.get_percent(), 42)`.

I can make it more explicit like `assert_eq!(flush_stop_evt.get_reset_time(), true)` if you prefer.

gstreamer/src/message.rs
fengalin: Yes indeed... I can do the change in GStreamer if you want me to.

gstreamer/src/event.rs
sdroege: Oh it's a boolean, not a time :) right. That's all OK then

gstreamer/src/message.rs
sdroege: That would be great :) let's try to get that in before 1.14

fengalin: All right!

fengalin: Just a comment: I don't think it is an issue using `ffi::gst_message_get_structure` in the `build` method. There is no extra fields at this point. I think it is better to use `ffi::gst_message_get_structure` and `ffi::gst_event_get_structure` in the `build`s as they don't allocate `structure`s when they are not needed. We should add `get_writable_structure` for both `gst::Message` and `gst::Event` for the user to add extra fields after the build.

sdroege: They can only be safely added during the build. Otherwise you need to ensure that the refcount is 1 for the message to be writable, etc

I'd prefer to have messages and events immutable after creation. We can prevent the allocation by not calling anything if no other fields

fengalin: I see! Let me a couple of minutes to do that too.

fengalin: BTW: https://bugzilla.gnome.org/show_bug.cgi?id=792928

gstreamer/src/event.rs
```diff
@@ -1667,3 +1667,50 @@ impl<'a> CustomBothOobBuilder<'a> {
...
+        }
+
+        // Event with arguments
+        let flush_stop_evt = Event::new_flush_stop(true)
+            .other_fields(&[("extra-field", &true)])
+            .build();
+        match flush_stop_evt.view() {
+            EventView::FlushStop(flush_stop_evt) => {
+                assert_eq!(flush_stop_evt.get_reset_time(), true);
+                assert!(flush_stop_evt.0.get_structure().is_some());
```
fengalin: What would you think of implementing a `deref` on the concret events & messages in order to get direct access to the `EventRef` & `MessageRef` methods without using `.0`?

gstreamer/src/message.rs
sdroege: Thanks, merged that one

gstreamer/src/event.rs
sdroege: Yes, I thought I did that already. `.0` should also be private.

Please add that if you want, same for event and query if needed :)

gstreamer/src/message.rs
```diff
@@ -1183,13 +1183,17 @@ macro_rules! message_builder_generic_impl {
...
-                    let s = StructureRef::from_glib_borrow_mut(
-                        ffi::gst_message_get_structure(msg) as *mut _
-                    );
-
-                    for (k, v) in self.other_fields {
-                        s.set_value(k, v.to_send_value());
+                if !self.other_fields.is_empty() {
+                    // TODO: add call ffi::gst_message_writable_structure instead
+                    // when https://bugzilla.gnome.org/show_bug.cgi?id=792928 is merged
+                    // and matching feature is activated
```
sdroege: For the 0.11 release, we should make the other-fields API conditional on the `v1_14` feature (and not have it at all for older versions). But that's for when 1.14 is actually release, in 1-2 months or so

gstreamer/src/event.rs
sdroege: Also sorry for not replying anything on the TOC bug so far, it needs some harder thinking by Tim and myself and we didn't get to it yet. If you're coming to FOSDEM, maybe we can discuss there?

It's on my list in any case

fengalin: OK! Thank you for the feedback. I'm not coming unfortunately. I can open a thread on the mailing list if that helps.

sdroege: Not really, it's mostly a matter of finding time to discuss and think it through :)

gstreamer/src/message.rs
```diff
@@ -1230,7 +1214,7 @@ macro_rules! message_builder_generic_impl {
 pub struct EosBuilder<'a> {
     src: Option<Object>,
     seqnum: Option<Seqnum>,
-    other_fields: Vec<(&'a str, &'a ToSendValue)>,
+    #[allow(unused)] other_fields: Vec<(&'a str, &'a ToSendValue)>,
```
fengalin: I think we should factorize the attributes `other_fields` in a generic struct `MessageBuilder`.

gstreamer/src/event.rs
```diff
@@ -494,28 +501,50 @@ pub enum EventView<'a> {
...
 
-pub struct FlushStart<'a>(&'a EventRef);
+macro_rules! declare_concrete_event(
+    ($name:ident) => {
+        pub struct $name<'a> {
+            inner: &'a EventRef
+        }
+
+        impl<'a> $name<'a> {
+            pub fn new(inner: &'a EventRef) -> Self {
```
sdroege: This should not be public. Also why do you change it from a tuple strict? :)

gstreamer/src/message.rs
```diff
@@ -1165,14 +1189,7 @@ macro_rules! message_builder_generic_impl {
             }
         }
 
-        pub fn other_fields(self, other_fields: &[(&'a str, &'a ToSendValue)]) -> Self {
```
sdroege: For the 0.10 branch this should stay, do best to remove it in a separate commit in this PR

gstreamer/src/message.rs
sdroege: Ok

gstreamer/src/query.rs
```diff
@@ -242,14 +242,56 @@ pub enum QueryView<T> {
...
+            }
+        }
+
+        impl<'a> $name<&'a mut QueryRef> {
+            pub fn get_mut_query(&mut self) -> &mut QueryRef {
+                self.inner
+            }
+        }
+
+        impl<'a> Deref for $name<&'a mut QueryRef> {
```
sdroege: Also DerefMut

gstreamer/src/event.rs
sdroege: Note that .0 was private before too. You could only use it in the test because it is in a sub module

fengalin: I couldn't find another way to make the inner `EventRef` private.

fengalin: Ah OK! Will revert this then. :)

fengalin:
I have reworked the patch set as it was getting messy. The removal of method `other_fields` in now in a separate commit. I hope this is ok now.

I wanted to go further in the rework of concrete builders using `default` in order to simplify `new` methods, but it was not as straight forward as I would have liked. I'll probably try it in a separate PR sometimes.

gstreamer/src/message.rs
```diff
@@ -1171,13 +1171,7 @@ impl<'a> MessageBuilder<'a> {
...
     }
 
-    fn clone_and_chain(&self, other_fields: &[(&'a str, &'a ToSendValue)]) -> Self {
-        Self {
-            other_fields: self.other_fields.iter().cloned()
-                .chain(other_fields.iter().cloned())
-                .collect()
-        }
-    }
+    // TODO: restore clone_and_chain method and condition it to the "v1_14" feature
```
sdroege: I would just comment it out for now instead of removing. That way adding it back in a few weeks is going to be less annoying :)

sdroege:
The diff is too big to comment on it in GitHub apparently. So for the second commit:

- If you factor out things into a generic EventBuilder, there's not only the other fields but also the seqnum and running time offset that is generic to all of them. But apart from having the common things in another struct, this does not really seem to simplify anything or does it?
- Same thing for MessageBuilder, but there it's src and seqnum
- Also you implemented a small behaviour difference (that makes sense!): calling other_fields multiple times on the builder is now merging the old and new other_fields
- DerefMut on the Query struct seems suboptimal. What we really want to return is a `&mut QueryRef`, not a `&mut QueryView`. See the previous `get_mut_query` functions

fengalin:
> But apart from having the common things in another struct, this does not really seem to simplify anything or does it?

Yeah, I was a bit hesitant after doing it. The way I see it, it gets the `Vec` full type declaration and logic in one place and leaves trivial attribute declarations and initializations to the concrete implementations. With the `default` thing, only the declaration would be required, removing repeated not so important stuff. Beside, I though that showing that these attributes were generic to all events / messages helped understand the architecture. With the extra attributes you mentioned, I think it makes even more sense. What do you prefer? Shall I revert to each concrete implementation holding these or can I go on with the generic `struct`s?

> calling other_fields multiple times on the builder is now merging the old and new other_fields

I missed that one, but it's great! It could help people implement conditionally additive extra-fields elegantly. And it's on par with what we can do using the C API.

> DerefMut on the Query struct seems suboptimal

OK, give me some time to sort this out. I think I will get the Query stuff apart, maybe in another PR, so that it doesn't hold this one.

sdroege:
> Yeah, I was a bit hesitant after doing it. The way I see it, it gets the Vec full type declaration and logic in one place and leaves trivial attribute declarations and initializations to the concrete implementations. With the default thing, only the declaration would be required, removing repeated not so important stuff. Beside, I though that showing that these attributes were generic to all events / messages helped understand the architecture. With the extra attributes you mentioned, I think it makes even more sense. What do you prefer? Shall I revert to each concrete implementation holding these or can I go on with the generic structs?

I would move the common fields into a common struct, yes. And then with adding a `Default` impl it probably shortens the code and reduces the noise and repetition.

> OK, give me some time to sort this out. I think I will get the Query stuff apart, maybe in another PR, so that it doesn't hold this one.

I also don't have a solution for that one. IIRC adding the duplicated getter functions everywhere was my solution and I tried using Deref/DerefMut first but it ended up being all weird.

The Query stuff is all good *except* for the Deref impls. So maybe only move the Deref/DerefMut part into a different PR?

sdroege:
So I guess to move forward here, let's move out the Query Deref/DerefMut stuff into another PR. And then either only solve the initial problem with the argument-less messages here, or feel free to also at the same time (like in the second commit) clean up all the repetition and in addition move the seqnum/src/etc common fields into the new struct and add a `Default` impl. (Currently I wouldn't merge that one yet as it feels half-finished without also moving the other common fields)

I'm happy with all those changes in any case, thanks a lot :)

fengalin:
Done! I'll do the other PRs tomorrow. :)

PR #76, opened 2018-01-28
Event, Message & Query: allow dereferencing concrete types to access generic type methods

fengalin:
Implement `deref` for concrete derivatives of `Event`, `Message` and `Query`. This allows accessing generic methods such as `is_sticky` for events or `get_src` for messages and `get_structure` for all.

This is also an opportunity to factorize common attributes in generic builders for events and messages.

PR #77, opened 2018-01-29
Backport PR #75 & #76 to 0.10

fengalin:
See https://github.com/sdroege/gstreamer-rs/pull/75 & https://github.com/sdroege/gstreamer-rs/pull/76

### DIFF
--- a/examples/src/bin/appsink.rs
+++ b/examples/src/bin/appsink.rs
@@ -137,7 +137,7 @@ fn main_loop(pipeline: gst::Pipeline) -> Result<(), Error> {
             MessageView::Error(err) => {
                 pipeline.set_state(gst::State::Null).into_result()?;
                 Err(ErrorMessage {
-                    src: msg.get_src()
+                    src: err.get_src()
                         .map(|s| s.get_path_string())
                         .unwrap_or_else(|| String::from("None")),
                     error: err.get_error().description().into(),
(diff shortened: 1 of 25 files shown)


fengalin:
Any other branches you want to backport to?

sdroege:
Only 0.10, thanks :) I would also have to check if there are any other changes to backport from master since last time I did that.

sdroege:
I might prefer to only take the first commit, it's a lot of churn to also backport the others, and it's mostly cleanup anyway.

sdroege:
I took that one commit now, and leave the others for 0.11/master :) That will be released soon enough anyway

Another thing for the future, please try to keep line lengths in commit messages. A bit more than 80 characters is ok, but not always and not that much :)

fengalin:
> Another thing for the future, please try to keep line lengths in commit messages. A bit more than 80 characters is ok, but not always and not that much :)

Sorry about that. My editor is configured for [rust-fmt defaults](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#max_width) for Rust. Maybe we should add a `rustfmt.toml` to enforce line width to 80. I can do that if you want me to.

sdroege:
Oh it's only about the commit messages. The code is fine, I use the rustfmt defaults too

fengalin:
' misread your comment. My bad!

PR #78, opened 2018-02-02
Query: proposal to solve the deref/deref_mut issue

fengalin:
While trying to solve the deref/deref_mut issue on concrete types, I came up with the approach in this commit. The idea is to allow working with concrete implementations directly. Except for pad probes, views are no longer necessary. See unit test and examples.

In pad probes, the user gets the query view directly. E.g.:
``` rust
    src_pad.add_probe(PadProbeType::QUERY_UPSTREAM, |_, probe_info| {
        if let Some(PadProbeData::Query(ref mut query_view)) = probe_info.data {
            match query_view {
                &mut QueryView::Duration(ref mut duration) => {
                    duration.set(2 * SECOND);
                },
                _ => (),
            }
        }
        PadProbeReturn::Ok
    });
```

Note: calling `view` consumes the query, but I don't think this is an issue as the idea is to work on concrete implementations and the user should no longer need to call `view` anyway.

If I'm not mistaken, this could apply to `Event`s & `Message`s too.

### DIFF
--- a/gstreamer/src/pad.rs
+++ b/gstreamer/src/pad.rs
@@ -183,7 +183,7 @@ pub trait PadExtManual {
 
     fn set_query_function<F>(&self, func: F)
     where
-        F: Fn(&Pad, &Option<::Object>, ::QueryView) -> bool + Send + Sync + 'static;
+        F: Fn(&Pad, &Option<::Object>, &mut ::QueryView) -> bool + Send + Sync + 'static;
 
     fn set_unlink_function<F>(&self, func: F)
     where
@@ -587,11 +587,11 @@ impl<O: IsA<Pad>> PadExtManual for O {
 
     fn set_query_function<F>(&self, func: F)
     where
-        F: Fn(&Pad, &Option<::Object>, ::QueryView) -> bool + Send + Sync + 'static,
+        F: Fn(&Pad, &Option<::Object>, &mut ::QueryView) -> bool + Send + Sync + 'static,
     {
         unsafe {
             let func_box: Box<
-                Fn(&Pad, &Option<::Object>, ::QueryView) -> bool + Send + Sync + 'static,
+                Fn(&Pad, &Option<::Object>, &mut ::QueryView) -> bool + Send + Sync + 'static,
             > = Box::new(func);
             ffi::gst_pad_set_query_function_full(
                 self.to_glib_none().0,
@@ -884,9 +884,9 @@ unsafe extern "C" fn trampoline_pad_probe(
                 )))
             } else if (*data).type_ == Query::static_type().to_glib() {
                 data_type = Some(Query::static_type());
-                Some(PadProbeData::Query(Query::to_view(
-                    data as *mut ffi::GstQuery,
-                )))
+                Some(PadProbeData::Query(
+                    Query::from_glib_none(data as *mut ffi::GstQuery).into()
+                ))
             } else if (*data).type_ == Event::static_type().to_glib() {
                 data_type = Some(Event::static_type());
                 Some(PadProbeData::Event(from_glib_none(
@@ -1121,7 +1121,7 @@ unsafe extern "C" fn trampoline_query_function(
     func(
         &from_glib_borrow(pad),
         &from_glib_borrow(parent),
-        ::Query::to_view(query),
+        ::Query::from_glib_none(query).into(),
     ).to_glib()
 }
 
--- a/gstreamer/src/pad.rs
+++ b/gstreamer/src/pad.rs
@@ -18,6 +18,9 @@ use SpecificFormattedValue;
 use FlowReturn;
 use Query;
 use QueryRef;
+// This is the generic `View` trait which would be moved in its own file
+// if this solution was adopted
+use query::View;
 use Event;
 use StaticPadTemplate;
 use miniobject::*;
@@ -183,7 +186,7 @@ pub trait PadExtManual {
 
     fn set_query_function<F>(&self, func: F)
     where
-        F: Fn(&Pad, &Option<::Object>, &mut ::Query) -> bool + Send + Sync + 'static;
+        F: Fn(&Pad, &Option<::Object>, &mut ::QueryView) -> bool + Send + Sync + 'static;
 
     fn set_unlink_function<F>(&self, func: F)
     where
@@ -587,11 +590,11 @@ impl<O: IsA<Pad>> PadExtManual for O {
 
     fn set_query_function<F>(&self, func: F)
     where
-        F: Fn(&Pad, &Option<::Object>, &mut ::Query) -> bool + Send + Sync + 'static,
+        F: Fn(&Pad, &Option<::Object>, &mut ::QueryView) -> bool + Send + Sync + 'static,
     {
         unsafe {
             let func_box: Box<
-                Fn(&Pad, &Option<::Object>, &mut ::Query) -> bool + Send + Sync + 'static,
+                Fn(&Pad, &Option<::Object>, &mut ::QueryView) -> bool + Send + Sync + 'static,
             > = Box::new(func);
             ffi::gst_pad_set_query_function_full(
                 self.to_glib_none().0,
@@ -882,11 +885,11 @@ unsafe extern "C" fn trampoline_pad_probe(
                 Some(PadProbeData::BufferList(from_glib_none(
                     data as *const ffi::GstBufferList,
                 )))
-            } else if (*data).type_ == QueryRef::static_type().to_glib() {
-                data_type = Some(QueryRef::static_type());
-                Some(PadProbeData::Query(
-                    from_glib_full(data as *mut ffi::GstQuery)
-                ))
+            } else if (*data).type_ == Query::static_type().to_glib() {
+                data_type = Some(Query::static_type());
+                Some(PadProbeData::Query(from_glib_borrow(
+                    data as *mut ffi::GstQuery
+                )))
             } else if (*data).type_ == Event::static_type().to_glib() {
                 data_type = Some(Event::static_type());
                 Some(PadProbeData::Event(from_glib_none(
@@ -923,7 +926,7 @@ unsafe extern "C" fn trampoline_pad_probe(
             }
         }
         None => {
-            assert_ne!(data_type, Some(QueryRef::static_type()));
+            assert_ne!(data_type, Some(Query::static_type()));
             if !(*info).data.is_null() {
                 ffi::gst_mini_object_unref((*info).data as *mut _);
                 (*info).data = ptr::null_mut();
@@ -1113,15 +1116,18 @@ unsafe extern "C" fn trampoline_query_function(
 ) -> glib_ffi::gboolean {
     let _guard = CallbackGuard::new();
     #[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
-    let func: &&(Fn(&Pad, &Option<::Object>, ::Query) -> bool
+    let func: &&(Fn(&Pad, &Option<::Object>, &mut ::QueryView) -> bool
                          + Send
                          + Sync
                          + 'static) = transmute((*pad).querydata);
 
+    let mut query: Query = from_glib_borrow(query);
+    assert!(query.is_writable());
+
     func(
         &from_glib_borrow(pad),
         &from_glib_borrow(parent),
-        from_glib_full(query),
+        query.make_mut(),
     ).to_glib()
 }
 
--- a/gstreamer/src/pad.rs
+++ b/gstreamer/src/pad.rs
@@ -17,8 +17,7 @@ use FormattedValue;
 use SpecificFormattedValue;
 use FlowReturn;
 use Query;
-use QueryRc;
-use QueryWrapper;
+use QueryRef;
 use Event;
 use StaticPadTemplate;
 use miniobject::*;
@@ -87,7 +86,7 @@ pub struct PadProbeInfo {
 pub enum PadProbeData {
     Buffer(Buffer),
     BufferList(BufferList),
-    Query(QueryWrapper),
+    Query(Query),
     Event(Event),
     Unknown,
 }
@@ -117,15 +116,15 @@ pub trait PadExtManual {
     fn pull_range(&self, offset: u64, size: u32) -> Result<Buffer, FlowReturn>;
     fn get_range(&self, offset: u64, size: u32) -> Result<Buffer, FlowReturn>;
 
-    fn peer_query(&self, query: &mut Query) -> bool;
-    fn query(&self, query: &mut Query) -> bool;
+    fn peer_query(&self, query: &mut QueryRef) -> bool;
+    fn query(&self, query: &mut QueryRef) -> bool;
     fn query_default<'a, P: IsA<Object> + 'a, Q: Into<Option<&'a P>>>(
         &self,
         parent: Q,
-        query: &mut Query,
+        query: &mut QueryRef,
     ) -> bool;
-    fn proxy_query_caps(&self, query: &mut Query) -> bool;
-    fn proxy_query_accept_caps(&self, query: &mut Query) -> bool;
+    fn proxy_query_caps(&self, query: &mut QueryRef) -> bool;
+    fn proxy_query_accept_caps(&self, query: &mut QueryRef) -> bool;
 
     fn event_default<'a, P: IsA<::Object> + 'a, Q: Into<Option<&'a P>>>(
         &self,
@@ -184,7 +183,7 @@ pub trait PadExtManual {
 
     fn set_query_function<F>(&self, func: F)
     where
-        F: Fn(&Pad, &Option<::Object>, &mut ::QueryWrapper) -> bool + Send + Sync + 'static;
+        F: Fn(&Pad, &Option<::Object>, &mut ::Query) -> bool + Send + Sync + 'static;
 
     fn set_unlink_function<F>(&self, func: F)
     where
@@ -312,7 +311,7 @@ impl<O: IsA<Pad>> PadExtManual for O {
         }
     }
 
-    fn query(&self, query: &mut Query) -> bool {
+    fn query(&self, query: &mut QueryRef) -> bool {
         unsafe {
             from_glib(ffi::gst_pad_query(
                 self.to_glib_none().0,
@@ -321,7 +320,7 @@ impl<O: IsA<Pad>> PadExtManual for O {
         }
     }
 
-    fn peer_query(&self, query: &mut Query) -> bool {
+    fn peer_query(&self, query: &mut QueryRef) -> bool {
         unsafe {
             from_glib(ffi::gst_pad_peer_query(
                 self.to_glib_none().0,
@@ -333,7 +332,7 @@ impl<O: IsA<Pad>> PadExtManual for O {
     fn query_default<'a, P: IsA<Object> + 'a, Q: Into<Option<&'a P>>>(
         &self,
         parent: Q,
-        query: &mut Query,
+        query: &mut QueryRef,
     ) -> bool {
         skip_assert_initialized!();
         let parent = parent.into();
@@ -347,7 +346,7 @@ impl<O: IsA<Pad>> PadExtManual for O {
         }
     }
 
-    fn proxy_query_accept_caps(&self, query: &mut Query) -> bool {
+    fn proxy_query_accept_caps(&self, query: &mut QueryRef) -> bool {
         unsafe {
             from_glib(ffi::gst_pad_proxy_query_accept_caps(
                 self.to_glib_none().0,
@@ -356,7 +355,7 @@ impl<O: IsA<Pad>> PadExtManual for O {
         }
     }
 
-    fn proxy_query_caps(&self, query: &mut Query) -> bool {
+    fn proxy_query_caps(&self, query: &mut QueryRef) -> bool {
         unsafe {
             from_glib(ffi::gst_pad_proxy_query_accept_caps(
                 self.to_glib_none().0,
@@ -588,11 +587,11 @@ impl<O: IsA<Pad>> PadExtManual for O {
 
     fn set_query_function<F>(&self, func: F)
     where
-        F: Fn(&Pad, &Option<::Object>, &mut ::QueryWrapper) -> bool + Send + Sync + 'static,
+        F: Fn(&Pad, &Option<::Object>, &mut ::Query) -> bool + Send + Sync + 'static,
     {
         unsafe {
             let func_box: Box<
-                Fn(&Pad, &Option<::Object>, &mut ::QueryWrapper) -> bool + Send + Sync + 'static,
+                Fn(&Pad, &Option<::Object>, &mut ::Query) -> bool + Send + Sync + 'static,
             > = Box::new(func);
             ffi::gst_pad_set_query_function_full(
                 self.to_glib_none().0,
@@ -883,10 +882,10 @@ unsafe extern "C" fn trampoline_pad_probe(
                 Some(PadProbeData::BufferList(from_glib_none(
                     data as *const ffi::GstBufferList,
                 )))
-            } else if (*data).type_ == Query::static_type().to_glib() {
-                data_type = Some(Query::static_type());
+            } else if (*data).type_ == QueryRef::static_type().to_glib() {
+                data_type = Some(QueryRef::static_type());
                 Some(PadProbeData::Query(
-                    QueryRc::from_glib_full(data as *mut ffi::GstQuery).into()
+                    from_glib_full(data as *mut ffi::GstQuery)
                 ))
             } else if (*data).type_ == Event::static_type().to_glib() {
                 data_type = Some(Event::static_type());
@@ -924,7 +923,7 @@ unsafe extern "C" fn trampoline_pad_probe(
             }
         }
         None => {
-            assert_ne!(data_type, Some(Query::static_type()));
+            assert_ne!(data_type, Some(QueryRef::static_type()));
             if !(*info).data.is_null() {
                 ffi::gst_mini_object_unref((*info).data as *mut _);
                 (*info).data = ptr::null_mut();
@@ -1114,7 +1113,7 @@ unsafe extern "C" fn trampoline_query_function(
 ) -> glib_ffi::gboolean {
     let _guard = CallbackGuard::new();
     #[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
-    let func: &&(Fn(&Pad, &Option<::Object>, ::QueryWrapper) -> bool
+    let func: &&(Fn(&Pad, &Option<::Object>, ::Query) -> bool
                          + Send
                          + Sync
                          + 'static) = transmute((*pad).querydata);
@@ -1122,7 +1121,7 @@ unsafe extern "C" fn trampoline_query_function(
     func(
         &from_glib_borrow(pad),
         &from_glib_borrow(parent),
-        QueryRc::from_glib_full(query).into(),
+        from_glib_full(query),
     ).to_glib()
 }
 
--- a/gstreamer/src/pad.rs
+++ b/gstreamer/src/pad.rs
@@ -17,7 +17,7 @@ use FormattedValue;
 use SpecificFormattedValue;
 use FlowReturn;
 use Query;
-use QueryRef;
+use QueryView;
 use Event;
 use StaticPadTemplate;
 use miniobject::MiniObject;
@@ -75,18 +75,18 @@ impl FromGlib<libc::c_ulong> for PadProbeId {
     }
 }
 
-pub struct PadProbeInfo<'a> {
+pub struct PadProbeInfo {
     pub mask: PadProbeType,
     pub id: PadProbeId,
     pub offset: u64,
     pub size: u32,
-    pub data: Option<PadProbeData<'a>>,
+    pub data: Option<PadProbeData>,
 }
 
-pub enum PadProbeData<'a> {
+pub enum PadProbeData {
     Buffer(Buffer),
     BufferList(BufferList),
-    Query(&'a mut QueryRef),
+    Query(QueryView),
     Event(Event),
     Unknown,
 }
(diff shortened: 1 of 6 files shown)


fengalin:
I can reproduce the issue shown by CI after a `cargo update` with and without my commit... 'will investigate this later.

sdroege:
Some changes have to be done because of the glib-sys/gobject-sys updates. The GValue data is now an union, and we can access the v_pointer field there directly

sdroege:
Otherwise these changes look very interesting indeed, I'll have to look in more detail later. And yes, the same should apply for event/message too

sdroege:
Also you will still need the query views everywhere, where you get a generic query. So e.g. in the pad query function (see Pad::set_query_function), the pad probes as you saw, or in the virtual methods for handling queries (in gst-plugin-rs). But there's not much we can do about that.

sdroege:
Hmm but it only really makes sense for queries. The others you only create and pass on, and from there you get it as a generic message/event and have to use the view anyway.

But for queries it makes sense

fengalin:
> Hmm but it only really makes sense for queries. The others you only create and pass on, and from there you get it as a generic message/event and have to use the view anyway.

Sure, views are still necessary. IMHO, using the same principle for Event & Message would be valuable. E.g. an `add_watch` on a pipeline bus could pass the `MessageView` as an argument to the callback instead of the generic `Message`, then the user uses a `match` directly to get the concrete implementation and doesn't have to worry about generic v.s. concrete messages since they can now `deref` to the generic implementation. This would be similar to the test for `Pad::set_query_function` mentioned in my last commit's message:
``` rust
    src_pad.set_query_function(|_pad, _parent, mut query_view| {
        match query_view {
            gst::QueryView::Duration(ref mut duration) => {
                duration.set(2 * gst::SECOND);
                true
            },
            // etc.
        }
    });
```

I think we can avoid returning to views from Event, Message & Query instances (i.e. we could remove the `view` method which I left in the `declare_concrete_query` macro).

I'll take a look at gst-plugins-rs.

fengalin:
About the last commit: it is probably a case where a `downcast` would make more sense. I will investigate this tomorrow.

sdroege:
Sounds like it might indeed also be useful for events/messages, that would simplify things a bit I guess.

I'll review the whole changes here later today

fengalin:
My position today is that we need a means to switch between generic and concrete types for the use cases of `gst-plugin-rs`. When the plugin is invoked from C, it will pass a generic Query. Either:

- We stick to the View approach in the Rust API (e.g. in `BaseTransformImpl::query`) and we handle it the same way as for `add_pad_probe` and `set_query_function`, but sooner or later we'll have to retrieve the generic query from the View, e.g. to invoke `ffi`. I don't think it would be a good idea to use a `match` and handle all the cases for this. So at least we would need a means to switch from the View to the generic implementation directly.
- Or we implement upcasting / downcasting, but then we have to deal with the MiniObject model around.

In the short term, I'm planning on exploring the followings:

1. Figure out the consequences if we'd used `GstRc<_ConcreteType_>` instead of having  `GstRc<_GenericType_>` a member of `_ConcreteType_`. This way, we may be able to transfer ownership by consuming the `GstRc<_GenericType_>` into a `GstRc<_ConcreteType_>` and handle mutability nicely. Additionally, we could implement `upcast/downcast_ref`. I foresee a possible use of a `Box` due to type not being `Sized`. Would that be acceptable? I guess there are heap allocations in many places already.
2. See if the approach from the first point could allow avoiding a dedicated `GstRc` and rely on Rust mechanisms only. I probably missed something here, but I feel that what we want to achieve with `GstRc` could be handled by the borrow checker, except for the `ffi` to Rust types functions of course.

fengalin:
There might be an even better way with `from` and `into`!

sdroege:
If you find a better way of doing the `GstRc` stuff that would be great. I experimented with that for weeks and this was the nicest that I came up with, which was actually safe and could cover all the use-cases :) Also I like the symmetry to `Arc`.

fengalin:
I might have been a little audacious about the replacement of `GstRc` :) If you spent weeks on it, I wouldn't pretend to find a better way!

I think the `QueryView` approach could be viable in the context of `gst-plugin-rs`. This is still very theoretical since I didn't try it in place. However, I implemented the use cases I was thinking about in the unit tests.

sdroege:
> I might have been a little audacious about the replacement of GstRc :) If you spent weeks on it, I wouldn't pretend to find a better way!

You're not me and might have different ideas, so don't feel like you don't even have to try :) I'm sure there's a nicer or more convenient way of doing all this.

gstreamer/src/query.rs
```diff
@@ -270,10 +270,154 @@ pub enum QueryView {
...
+            }
+        }
+
+        impl Into<QueryView> for $name {
+            fn into(self) -> QueryView {
+                QueryView::$name($name(self.0))
+            }
+        }
+
+        impl Into<Query> for $name {
```
sdroege: Usually you would implement `From`, not `Into`. It's slightly more general AFAIU, and you can implement it here

gstreamer/src/pad.rs
```diff
@@ -116,15 +116,15 @@ pub trait PadExtManual {
     fn pull_range(&self, offset: u64, size: u32) -> Result<Buffer, FlowReturn>;
     fn get_range(&self, offset: u64, size: u32) -> Result<Buffer, FlowReturn>;
 
-    fn peer_query(&self, query: &mut QueryRef) -> bool;
-    fn query(&self, query: &mut QueryRef) -> bool;
+    fn peer_query(&self, query: &mut Query) -> bool;
+    fn query(&self, query: &mut Query) -> bool;
```
sdroege: With all these functions we now have the problem that it's a bug to pass a query in there that has a refcount > 1, but it's possible now. The query must be writable, which was the reason for passing a `&mut QueryRef` before (which guarantees exactly this).

gstreamer/src/pad.rs
```diff
@@ -75,18 +75,18 @@ impl FromGlib<libc::c_ulong> for PadProbeId {
...
-    pub data: Option<PadProbeData<'a>>,
+    pub data: Option<PadProbeData>,
 }
 
-pub enum PadProbeData<'a> {
+pub enum PadProbeData {
     Buffer(Buffer),
     BufferList(BufferList),
-    Query(&'a mut QueryRef),
+    Query(QueryView),
```
sdroege: Similarly here, this is guaranteed to be writable

gstreamer/src/value.rs
```diff
@@ -560,7 +560,7 @@ impl<'a> From<&'a [glib::SendValue]> for Array<'a> {
 
 impl<'a> FromValue<'a> for Array<'a> {
     unsafe fn from_value(v: &'a Value) -> Self {
-        let arr = (*v.to_glib_none().0).data[0] as *const glib_ffi::GArray;
+        let arr = (*v.to_glib_none().0).data[0].v_pointer as *const glib_ffi::GArray;
```
sdroege: This part is already merged but still shown in the PR. Maybe just drop the commit next time you rebase :)

tutorials/src/bin/basic-tutorial-4.rs
```diff
@@ -128,19 +128,14 @@ fn handle_message(custom_data: &mut CustomData, msg: &gst::GstRc<gst::MessageRef
...
-                        }
-                        _ => unreachable!(),
+                let mut seeking = gst::Query::new_seeking(gst::Format::Time);
+                if custom_data.playbin.query(&mut seeking) {
+                    let (seekable, start, end) = seeking.get_result();
+                    custom_data.seek_enabled = seekable;
+                    if seekable {
+                        println!("Seeking is ENABLED from {:?} to {:?}", start, end)
+                    } else {
+                        println!("Seeking is DISABLED for this stream.")
```
sdroege: This is a very nice usability improvement

gstreamer/src/query.rs
```diff
@@ -219,73 +327,123 @@ impl ToOwned for QueryRef {
...
+    Scheduling(Scheduling),
+    AcceptCaps(AcceptCaps),
+    Caps(Caps),
+    Drain(Drain),
+    Context(Context),
+    Other(Other),
     __NonExhaustive,
 }
 
+impl From<Query> for QueryView {
```
sdroege: With this you can take a `Query` with refcount > 1, get a `QueryView`, then call `&mut QueryView` functions. And those would fail. Or am I missing something?

fengalin: I think you're right. I'll look into that tomorrow.

sdroege: Thanks :) Generally this all looks very good usage-wise, but we need to be careful to not break things and to allow code that should not be able to compile

sdroege:
Also I'm wondering if we shouldn't go with downcast/upcast functions here. That's what it basically is

fengalin:
Yes, I wanted to explore that too.

fengalin:
These are the results of my analysis of current architecture (the one in master) for minobject based abstractions. The problem to address is tricky IMHO, so I may not have grasped all its complexity. I started experimenting in order to understand the motivations for current architecture and that lead me to some modifications. The commits in this PR implement a preview of the proposals in the second part. Several enhancements are still possible, but current status should permit its evaluation.


# Current Architecture

## Overview

`gstreamer-rs` encapsulates the C `MiniObject`'s behaviour in `GstRc<T>`. `GstRc<T>` controls the mutability of the underlying miniobject. If the miniobject `ref_count` is > 1, it can't be switched to mutable.

GStreamer's queries are based on the `MiniObject` model. Concrete query variantes share common behaviours but use additional values. In order to handle the variantes, `gstreamer-rs` uses the enum `QueryView`. `QueryView` allows downcasting a generic `ffi::GstQuery` received from the C code to the matching concrete Rust query. For the concrete query to respect the mutability constraints of the miniobject, it owns a `ref` or `ref mut` to the `GstRc<Query>`. The `QueryView` itself is obtained as a `ref` or `ref mut` so as to retain ownership on the concrete query and to restrain its mutability when applicable.

## Constraints related to using `MiniObject`s in current architecture

To access the query's miniobject as `ref mut` from a mutable `GstRc<Query>`, a developper needs to use either:

- `make_mut` which returns a `ref mut` to the original miniobject if it is writable (i.e. if it is not borrowed elsewhere), otherwise the miniobject is replaced with a copy and a `ref mut` to the miniobject is returned.
- `get_mut` which returns an option with a `ref mut` to the miniobject if it is writable or `None` otherwise.

When using a `GstRc<Query>` or a concrete query instance, the Rust object's mutability may not reflect the underlying miniobject's mutability. E.g.:

``` rust
    [...]
    match query.view_mut() {
        QueryView::Position(ref mut p) => {
            // Can access `p` as mutable here
            p.set(2 * ::SECOND);

            let mut structure = p.get_mut_structure();
                                ^
=> ERROR: cannot borrow immutable borrowed content as mutable
            [...]
        }
        [...]
    }
```

The concrete query `p` is mutable, however accessing the `structure`, which is part of the generic query, requires to explicitely obtain a `ref mut` on the generic object using `get_mut().unwrap()` (or `make_mut` which is guaranteed to keep the original miniobject here since the concrete query was obtained via a `ref mut QueryView`).

In the following example, a query with type `Position` is created within the Rust code, owned and marked as `mut`, however the query can't be `deref_mut` to a `&mut QueryRef` without using `make_mut`:

``` rust
    let mut p = Query::new_position(::Format::Time);
    let q: &mut QueryRef = &mut p; // or calling an `f(&mut QueryRef)`
                           ^^^^^^
=> ERROR: cannot borrow immutable borrowed content as mutable
```

In order to retain ownership and mutabilty, concrete queries can only be obtained from a `QueryView` via the `QueryRef::view()` or `QueryRef::view_mut()` functions. The `Query::new_...` functions which allow building queries with a specific query type return a generic `GstRc<QueryRef>`. This means that we can't use the concrete functions on the newly created instance:

``` rust
    let mut p = Query::new_position(::Format::Time);
    p.set(2 * ::SECOND);
      ^^^
=> ERROR: no method named `set` found for type `miniobject::GstRc<query::QueryRef>` in the current scope
```

A `QueryView` must be obtained, which will allow accessing the concrete variante.


# Proposal for a modified architecture

## Description

I will use `Query` since it is the object of this PR. The same should be applicable to other abstractions based on `GstRc<T>`.

The proposal is based on the following hypothesis:

- If the query was received from another language which uses the C miniobject model, the `writability` of the incoming query must be checked and access to it must be controlled, just like in current architecture.
- If the query was created in Rust, the borrow checker can handle ownership and mutability.

A query received as a C miniobject is encapsulated in its concrete type which is owned by a `QueryView` which is wrapped in a `QueryWrapper`. The `QueryWrapper` controls how the `QueryView` can be obtained, just like `QueryRef` in current architecture. Ownership and writabilty of the query are preserved. For ease of use, the `QueryWrapper` can be `deref` to serve as a generic `&Query`.

The `QueryView` behaves like in current architecture, except that the `QueryWrapper` owns it instead of the `QueryRef`. For ease of use, the `QueryView` can be `deref` to serve as a generic `&Query` or `&mut Query` if the `QueryWrapper` allowed the user to obtain a `&mut QueryView`.

Concrete queries can `deref` to the generic `&Query` or `&mut Query` in the same way as for `QueryView`. Concrete queries own their generic query, which means that a concrete query created from the Rust code will be able to access the generic functions. A standalone concrete query ownership can be trasnfered to a `QueryWrapper` or a `QueryView` if necessary.

## Examples

See the unit tests of `query.rs` for identified use cases. Here are a few extracts:

``` rust
    fn check_mut_view(view: &mut QueryView) {
        match *view {
            QueryView::Position(ref mut p) => {
                p.set(3 * ::SECOND);

                // `get_mut_structure` can be called
                // directly since `p` is `ref mut`
                p.get_mut_structure()
                    .set("check_mut", &true);
            }
            [...]
        }
    }
```

``` rust
        let mut p = Query::new_position(::Format::Time);

        // access generic function
        assert!(!p.is_serialized());

        // use specific setter
        p.set(2 * ::SECOND);
```

## Name changes

`GstRc<QueryRef>` is now aliased as `QueryRc` (previously `Query`) as it is used internally. Its role as a mutability controller is mainly handled by `QueryWrapper` in this architecture. `QueryRef` is now named `Query` as it appears in most of the methods.

## Limitations of the modified architecture

Identified limitations include:

- Increased code base. There are multiple `match` clauses to handle `deref` and `into` of the QueryView. A segment could be avoided if `GstRc<T>` implemented `deref_mut`. If we consider that the wrapper is responsible for checking writability, this would be possible.

- Since the ownership model is reversed (`QueryWrapper` owns `QueryView` owns `ConcreteQuery` owns `Query` instead of `Query` owns `QueryView` owns `ConcreteQuery`), it is no longer possible to build a `QueryView` from a function which received a `&Query` or `&mut Query`. A solution would be to use `&QueryView` or `&mut QueryView` as an argument to a function which may need to deal with concrete query variantes and use `&Query` and `&mut Query` for functions which deal with generic queries, such as those which pass queries to the C code. Another solution would to use `QueryWrapper` everywhere since they give access to everything. We could then pass concrete queries using `into()`.

## Possible evolutions

If this architecture is applicable to the other miniobject based abstraction, a generic Wrapper could be implemented. This wrapper could handle some (most?) of the role from GstRc<T>. This would also make the Wrapper the exclusive entry point for minobjects coming from C.

sdroege:
I didn't look in detail yet, that's for tomorrow. But the reason of GstRc is that it has the same API as Arc (also has get/make_mut that works the same), so it should feel somewhat familiar to Rust developers and not a completely new concept.

I'll check the details of your changes tomorrow, thanks :)

gstreamer/src/query.rs
```diff
@@ -15,188 +15,188 @@ use std::ptr;
...
 use std::ffi::CStr;
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 
 use glib;
 use glib::translate::{from_glib, from_glib_full, ToGlib, ToGlibPtr};
 
 #[repr(C)]
-pub struct QueryRef(ffi::GstQuery);
+pub struct Query(ffi::GstQuery);
```
sdroege: Also generally these things are currently called `XXXRef` and the owned thing is called `XXX`. See `Structure`, `VideoFrame`. Idea here being similar to having `Path` and `PathBuf`, consistently named.

gstreamer/src/query.rs
```diff
@@ -15,188 +15,188 @@ use std::ptr;
...
+    pub fn new_context(context_type: &str) -> Context {
+        assert_initialized_main_thread!();
+        Context(
+            unsafe { from_glib_full(ffi::gst_query_new_context(context_type.to_glib_none().0)) }
+        )
     }
 }
 
-impl glib::types::StaticType for QueryRef {
+impl glib::types::StaticType for Query {
```
sdroege: With this change it would be possible to store queries in GValues, but that can probably be added again somehow

gstreamer/src/query.rs
```diff
@@ -208,84 +208,273 @@ impl fmt::Debug for QueryRef {
...
+            _ => unimplemented!("Converting QueryView into Query for unknown query type"),
+        }
+    }
+}
+
+// Bypass `QueryRc`: if `QueryView` is a `&mut`, this means that
+// `Query` can also be accessed as `&mut` thanks to the controls
+// enforced by `QueryWrapper`
+impl DerefMut for QueryView {
+    fn deref_mut(&mut self) -> &mut Query {
```
sdroege: I believe with `impl From<QueryRc> for QueryView` you can get a `QueryView` from every `GstRc<Query>`, and with the `deref_mut` here you could always get a `&mut Query`, no matter what the current refcount is

gstreamer/src/query.rs
sdroege: generally, I'm not too happy with the naming but we can fix that later :)

gstreamer/src/pad.rs
```diff
@@ -311,7 +312,7 @@ impl<O: IsA<Pad>> PadExtManual for O {
         }
     }
 
-    fn query(&self, query: &mut QueryRef) -> bool {
+    fn query(&self, query: &mut Query) -> bool {
```
sdroege: Note that the same pattern does not work for events/messages/buffers/etc, as you pass those by value instead of borrowing them (mutably). That's the special part about queries here

gstreamer/src/query.rs
```diff
@@ -208,84 +208,273 @@ impl fmt::Debug for QueryRef {
...
+    }
+}
+
+impl From<QueryView> for QueryWrapper {
+    fn from(query_view: QueryView) -> Self {
+        QueryWrapper(query_view)
+    }
+}
+
+pub struct QueryWrapper(QueryView);
```
sdroege: What's the actual difference between a `QueryWrapper` and a `GstRc<QueryRef>`? The below is basically the `Arc` API. And from it you can get the view and the `&QueryRef`.

I don't understand why those two types can't be the same, I'm probably missing something

gstreamer/src/query.rs
```diff
@@ -15,188 +15,188 @@ use std::ptr;
...
+        assert_initialized_main_thread!();
+        let filter = filter.into();
+        Caps(
+            unsafe { from_glib_full(ffi::gst_query_new_caps(filter.to_glib_none().0)) }
+        )
     }
 
-    pub fn view(&self) -> QueryView<&Self> {
-        let type_ = unsafe { (*self.as_ptr()).type_ };
+    pub fn new_drain() -> Drain {
```
sdroege: That the constructors return the concrete types seems useful to me in any case. And what is needed for these is that they can be somehow converted to a `GstRc<QueryRef>`, and that you can borrow (possibly mutably, that's why the `QueryView` had a type parameter before) a `&QueryRef` from it to pass it around.

So `view` on `GstRc<QueryRef>` could return a immutable `QueryView` somehow that allows to get an immutable `Drain`, etc (so we need some more type parameters). And `view_mut` on `GstRc<>` would give you the mutable variants. And that type parameter could probably be `&mut QueryRef` or `&QueryRef` like before?

I think with that we would cover all use cases without changing everything and without going away from consistency with the other types.

gstreamer/src/query.rs
fengalin: Ok, I'll revert this.

gstreamer/src/query.rs
fengalin: That's true because `GstRc<Query>` is still in charge of creating the `QueryRef` from the `ffi::GstQuery`. But, if we consider the `QueryWrapper` is now in charge, we could avoid this. See below for the discussion about the `QueryWrapper` v.s. `GstRc<Query>`.

I will update the code to be more explcit.

gstreamer/src/pad.rs
fengalin: In that case, we could implement something like `try_to_owned` that would transfer ownership of the `xxView` when `is_writable` is `true`.

gstreamer/src/query.rs
fengalin: In my perspective, `QueryWrapper` is acting as `GstRc<T>`. The differences are:

- The position in the architecture: `QueryWrapper` is at the top level of the ownership chain, when `GstRc<T>` can be owned by a concrete type which can be owned by a `QueryView`.
- `GstRc<T>`'s parameter `T` is expected to be tied to a `ffi::type` which prevented me from using it to host a `QueryView`.

I kept `GstRc<T>` in the composition chain as a convenience for reusing existing functions. I should clear that by implementing them on the `QueryWrapper` and get it out of the way. Clearly, they provide the same functionalities. If we adopted this approach, we could use `GstRc<T>` in place of the `QueryWrapper`. I just didn't want to change all the code of `gstreamer-rs` for this architecture preview :)

gstreamer/src/query.rs
fengalin: I can give this a try, but isn't this what the `QueryWrapper` does? :)

One thing I prefer in my proposal is that the composition chain is more linear:
`QueryWrapper` owns `QueryView` owns `ConcreteQuery` owns `QueryRef`.

If I'm not mistaken, in current architecture, we have:
`QueryRef` owns `QueryView` owns `ConcreteQuery` owns a `ref` or `ref mut` to the initial `QueryRef`.

I had hard times understanding why it was designed that way.

sdroege: It's not so much about "owning" but about having another data structure that borrows from the other and allows accessing it in specific ways.

Current architecture is that `GstRc<QueryRef>` owns the query. You can borrow a `QueryRef` from it, mutably or not depending on refcount. If you have a `&QueryRef`, then you view into it (possibly mutably) which gives you access to the concrete content of the query (and the `&QueryRef` again for being able to modify that.

I think what could be done is that the concrete query views are made generic, so that you can get them by looking at a `&QueryRef` (possibly mutably) and also have them independently and then *owning* a `GstRc<QueryRef>` to which then can be transformed, and allow borrowing the `&QueryRef` (and mut variant) from it.

I.e. we would have something like `ConcreteQueryView<GstRc<QueryRef>>>`, `ConcreteQueryView<&mut QueryRef>` and `ConcreteQueryView<&QueryRef>`. I believe the first is only ever needed when we directly created a query and know that it has refcount==1, so that does not need to distinguish mutability.

sdroege: In any case, you're on to something and we'll find a nice solution for it all :)

fengalin:
I think I addressed the concerns you raised, except of course for the fact that this is a different architecture.

I also spent some times trying to figure out how to adapt current architecture in order to be able to instantiate concrete queries, but I'm facing issues dealing with the variations on the `Concrete<&QueryRef>`, `Concrete<&mut QueryRef>` and `Concrete<QueryRef>`. I may try again later and would submit another PR if I find a way. But I still think this approach does the job, and IMHO it's nice :)

gstreamer/src/pad.rs
```diff
@@ -183,7 +183,7 @@ pub trait PadExtManual {
 
     fn set_query_function<F>(&self, func: F)
     where
-        F: Fn(&Pad, &Option<::Object>, &mut ::QueryRef) -> bool + Send + Sync + 'static;
+        F: Fn(&Pad, &Option<::Object>, &mut ::Query) -> bool + Send + Sync + 'static;
```
sdroege: &mut Query has no meaning, as it's a refcount wrapper. You can always just clone it (increase refcount), and then you can get a mutable reference from it. Just that the inner type will have a refcount > 1 and is not mutable.

That's exactly the reason why before a &mut QueryRef was passed, a non-refcounted, borrowed reference, that statically has mutability guaranteed and can't in any case be cloned (or otherwise get the refcount increased).

In your design this should probably be a &mut QueryView?
Similar thoughts for Pad::query(), etc

gstreamer/src/pad.rs
```diff
@@ -882,11 +882,11 @@ unsafe extern "C" fn trampoline_pad_probe(
...
                 )))
-            } else if (*data).type_ == Query::static_type().to_glib() {
-                data_type = Some(Query::static_type());
-                Some(PadProbeData::Query(QueryRef::from_mut_ptr(
-                    data as *mut ffi::GstQuery,
-                )))
+            } else if (*data).type_ == QueryRef::static_type().to_glib() {
+                data_type = Some(QueryRef::static_type());
+                Some(PadProbeData::Query(
+                    from_glib_full(data as *mut ffi::GstQuery)
```
sdroege: This is wrong: we only borrow the query here, we don't own it. And it must be writable

gstreamer/src/pad.rs
```diff
@@ -923,7 +923,7 @@ unsafe extern "C" fn trampoline_pad_probe(
             }
         }
         None => {
-            assert_ne!(data_type, Some(Query::static_type()));
+            assert_ne!(data_type, Some(QueryRef::static_type()));
```
sdroege: Query should probably proxy the static type trait

gstreamer/src/pad.rs
```diff
@@ -1113,15 +1113,15 @@ unsafe extern "C" fn trampoline_query_function(
...
+    let func: &&(Fn(&Pad, &Option<::Object>, ::Query) -> bool
                          + Send
                          + Sync
                          + 'static) = transmute((*pad).querydata);
 
     func(
         &from_glib_borrow(pad),
         &from_glib_borrow(parent),
-        ::QueryRef::from_mut_ptr(query),
+        from_glib_full(query),
```
sdroege: This is wrong, same problem as with the pad probe. We don't own the query, we only borrow it mutably.

gstreamer/src/query.rs
```diff
@@ -15,114 +15,88 @@ use std::ptr;
...
 
-unsafe impl Send for QueryRef {}
-unsafe impl Sync for QueryRef {}
+    fn is_writable(&self) -> bool;
+    fn as_ref<'a>(&'a self) -> &'a Self::Generic;
 
-pub type Query = GstRc<QueryRef>;
+    // If the `View` is a `&mut`, this means that `Generic`
+    // can also be accessed as `&mut` thanks to the controls enforced by `GstRc2`
+    fn as_mut<'a>(&'a mut self) -> &'a mut Self::Generic;
```
sdroege: The lifetime parameter is probably unneeded in both cases here

gstreamer/src/query.rs
```diff
@@ -15,114 +15,88 @@ use std::ptr;
...
-    pub fn new_convert<V: Into<GenericFormattedValue>>(value: V, dest_fmt: ::Format) -> Self {
-        assert_initialized_main_thread!();
-        let value = value.into();
-        unsafe {
-            from_glib_full(ffi::gst_query_new_convert(
-                value.get_format().to_glib(),
-                value.get_value(),
-                dest_fmt.to_glib(),
-            ))
+    pub fn try_view_mut<'a>(&'a mut self) -> Option<&'a mut V> {
```
sdroege: This is more a get_mut, compare Arc/Rc. And a make_mut would be useful

gstreamer/src/query.rs
```diff
@@ -15,114 +15,88 @@ use std::ptr;
...
-    pub fn new_custom(structure: ::Structure) -> Self {
-        assert_initialized_main_thread!();
-        unsafe {
-            from_glib_full(ffi::gst_query_new_custom(
-                ffi::GST_QUERY_CUSTOM,
-                structure.into_ptr(),
-            ))
+    pub fn try_into_owned(self) -> Option<V::Generic> {
+        if self.0.is_writable() {
+            Some(self.0.into_owned())
```
sdroege: IntoOwned should always be possible. It creates a completely new query

gstreamer/src/query.rs
```diff
@@ -15,114 +15,88 @@ use std::ptr;
...
     }
+}
 
-    pub fn new_accept_caps(caps: &::Caps) -> Self {
-        assert_initialized_main_thread!();
-        unsafe { from_glib_full(ffi::gst_query_new_accept_caps(caps.as_mut_ptr())) }
-    }
+// *** Architecture Preview ***
+// Query specific definitions start here
+pub struct QueryRef(*mut ffi::GstQuery);
```
sdroege: This is always an owned pointer, or how would you distinguish between owned and borrowed here? Borrowing without owning is needed for the pad functions for example

Also I think this is unsafe. Whenever you provide a &mut QueryRef now, it would be possible to replace it. But that would mean that the pointer would be replaced, potentially causing interesting memory unsafety issues with regard to lifetimes of the underlying object.
That's what the QueryRef (and the other Ref types) are the actual C types, so that getting a mutable reference would replace the underlying struct instead of the pointer.

gstreamer/src/query.rs
```diff
@@ -143,50 +117,140 @@ impl QueryRef {
...
-            ((*self.as_ptr()).type_ as u32) & (ffi::GST_QUERY_TYPE_SERIALIZED.bits()) != 0
+            ((*self.0).type_ as u32) & (ffi::GST_QUERY_TYPE_SERIALIZED.bits()) != 0
         }
     }
+}
 
-    pub fn view(&self) -> QueryView<&Self> {
-        let type_ = unsafe { (*self.as_ptr()).type_ };
+impl QueryRef {
+    pub fn new_position(fmt: ::Format) -> Position {
```
sdroege: These would make more sense on Query itself, naming-wise. Even if they don't return a literal query. They don't return a QueryRef either :)

fengalin:
I didn't generalize using `QueryView` as an argument everywhere yet (just used it in the `query_function` callback). I tried several approaches, but it seems that keeping the same usability as in current state would require providing implementations for `QueryView<&T>`, `QueryView<&mut T>`, `QueryView<T>`. So I guess that invalidates this approach :) I will try again later and also will try again on current architecture.

sdroege:
Well, that's more or less exactly what I was going to implement :)

fengalin:
Just noticed my comment was ambiguous. I meant to write: "I guess this invalidates *my* approach".

sdroege:
> Just noticed my comment was ambiguous. I meant to write: "I guess this invalidates my approach".

Ah don't worry, I misread your sentence even more the first time I read it. Now it all makes sense :) You look at implementing the other idea, or I keep it on my list?

fengalin:
I tried again this morning, but I couldn't find a way to avoid code duplication for concrete queries. I will try again later and formalize the issue if I'm still stuck on it.

fengalin:
This approach is abandoned in favour of: https://github.com/sdroege/gstreamer-rs/pull/82

PR #82, opened 2018-02-15
Query: concrete type instantiation

fengalin:
Allow instantiating and dereferencing concrete queries. The motivation for this proposal is to allow the following usability enhancements:

Concrete queries mutability guaranteed by the borrow checker, including for generic functions:
``` rust
    let mut p = Query::new_position(::Format::Time);
    p.get_mut_structure().set("check_mut", &true);
```

Concrete queries functions available in place:
``` rust
    let mut q = gst::Query::new_duration(gst::Format::Time);
    let duration = if pipeline.query(&mut q) {
        Some(q.get_result())
    } else {
        None
    };
```

### DIFF
--- a/gstreamer/src/query.rs
+++ b/gstreamer/src/query.rs
@@ -15,10 +15,10 @@ use std::ptr;
 use std::mem;
 use std::fmt;
 use std::ffi::CStr;
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 
 use glib;
-use glib::translate::{from_glib, from_glib_full, ToGlib, ToGlibPtr};
+use glib::translate::*;
 
 #[repr(C)]
 pub struct QueryRef(ffi::GstQuery);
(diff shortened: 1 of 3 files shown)


gstreamer/src/query.rs
```diff
@@ -15,114 +15,185 @@ use std::ptr;
...
-        assert_initialized_main_thread!();
-        unsafe { from_glib_full(ffi::gst_query_new_formats()) }
+// *** Draft Rework ***
+// Query specific definitions start here
+glib_wrapper! {
+    pub struct QueryRef(Shared<ffi::GstQuery>);
+
+    match fn {
+        ref => |ptr| ffi::gst_mini_object_ref(ptr as *mut ffi::GstMiniObject),
+        unref => |ptr| ffi::gst_mini_object_unref(ptr as *mut ffi::GstMiniObject),
```
sdroege: Due to this, you can take a `&mut QueryRef` (as from `get_mut` e.g.) and use its `Clone` impl to get a `QueryRef` that uses the very same pointer but it has a refcount > 1. You then have a `&mut QueryRef` that is not writable anymore.

gstreamer/src/query.rs
```diff
@@ -208,17 +273,156 @@ impl fmt::Debug for QueryRef {
     }
 }
 
-impl ToOwned for QueryRef {
-    type Owned = GstRc<QueryRef>;
+impl Query {
+    pub fn new_position(fmt: ::Format) -> Position<QueryRef> {
```
sdroege: My idea here was that we put `Query` into the type parameter directly. So that there can be `Position<Query>`, `Position<&QueryRef>` and `Position<&mut QueryRef>`. The first would own the query, and own it uniquely (refcount == 1), the others borrow the query and the last one has it borrowed uniquely (refcount == 1)

sdroege:
>  is it possible to receive foreign queries with a ref_count > 1 in query_function and pad_probe?

No, that would be a bug

> Did I break QueryRef writability logic by using the Shared model?

Yes, see comment :) But I think this is close to a full solution now nonetheless

gstreamer/src/query.rs
sdroege: I.e. `Position<GstRc<QueryRef>>` (`QueryRef` as it was before). You can then safely deref/deref_mut the inside of that to a `&QueryRef` and `&mut QueryRef` (because you know that refcount==1). And then wrap a `Position<&mut QueryRef>` around it, which gives access to all the API. Just not sure how to implement `Deref`/`DerefMut` so that we can directly go to a `&mut Position<&mut QueryRef>`, or how we otherwise get access to all the API we care about on the `Position<Query>` in a convenient way.

fengalin:
Thank you for your patience :)

gstreamer/src/query.rs
```diff
@@ -208,17 +114,182 @@ impl fmt::Debug for QueryRef {
     }
 }
 
-impl ToOwned for QueryRef {
-    type Owned = GstRc<QueryRef>;
+unsafe impl MiniObject for QueryRef {
```
sdroege: Can you move these things above `impl QueryRef` again to keep the diff more readable?

fengalin: Sure!

gstreamer/src/query.rs
```diff
@@ -260,12 +332,6 @@ macro_rules! declare_concrete_query(
             }
         }
 
-        impl<'a> $name<&'a mut QueryRef> {
```
sdroege: Why do you remove this?

gstreamer/src/query.rs
```diff
@@ -295,20 +382,20 @@ impl<'a> Position<&'a QueryRef> {
...
-            ffi::gst_query_parse_position(self.as_mut_ptr(), &mut fmt, ptr::null_mut());
+            ffi::gst_query_parse_position(self.0.as_mut_ptr(), &mut fmt, ptr::null_mut());
 
             from_glib(fmt)
         }
     }
 }
 
-impl<'a> Position<&'a mut QueryRef> {
+impl<T: AsMutPtr> Position<T> {
```
sdroege: I think we need to distinguish between things that can be mutable and things that can't. Right now this would allow to call `set()` on a `Position<&QueryRef>` or not? Because the `&mut self` is for `Position`, not `T`.

So this probably needs a `AsMutPtr` and `AsPtr` trait, or something like that

gstreamer/src/query.rs
```diff
@@ -208,17 +267,29 @@ impl fmt::Debug for QueryRef {
     }
 }
 
-impl ToOwned for QueryRef {
-    type Owned = GstRc<QueryRef>;
+pub unsafe trait AsMutPtr {
```
sdroege: This trait seems like a crutch, but I don't have a better suggestion so let's keep it like that and just not re-export it outside the query module :)

Maybe should be named `AsMutQueryPtr` or so, to make it clear from the name that this is about queries. OTOH if it's in the query module... so I guess just keep it all as is!

sdroege:
Just minor things left from what I can see :)

gstreamer/src/query.rs
fengalin: Yes, you're right.

This raises a concern though. The getters for the concrete types (e.g. `get_result`) use functions such as `ffi::gst_query_parse_position` which expect *mut ffi::GstQuery` as an argument. So either I cast them all as `*mut ffi::GstQuery` or I make `as_ptr` return a `*mut ffi::GstQuery`, which would seem contradictory in my opinion.

gstreamer/src/query.rs
fengalin: Ok. I wasn't too happy with it neither.

gstreamer/src/query.rs
sdroege: The GStreamer C API is not const-correct, like almost no C API :) Always return a mut pointer for both traits

sdroege: Or alternatively just cast the non-mut pointer to a mut one at the place of usage

fengalin: All right!

fengalin:
I'll squash the commits when you don't have more comments.

sdroege:
Looks all good to go :) Please squash the commits together and also fix up the commit message a bit to the latest version of reality (and try to not have lines much longer than 80 characters)

gstreamer/src/query.rs
```diff
@@ -260,32 +348,53 @@ macro_rules! declare_concrete_query(
...
-                    mem::transmute(self)
-                }
+                &self.0
+            }
+        }
+
+        impl DerefMut for $name<Query> {
+            fn deref_mut(&mut self) -> &mut Self::Target {
+                assert!(self.0.is_writable());
+                self.0.make_mut()
```
sdroege: This can (and probably should be) `self.0.make_mut().unwrap()`. It can't possibly fail, it does not need to go through `gst_query_make_writable()`.

sdroege: get_mut I mean, not make. Sorry

sdroege:
Thanks a lot for your patience, this improves things a lot :) I also updated gst-plugin-rs now

PR #83, opened 2018-02-16
Query: simplify concrete query declarations

fengalin:
Changes in https://github.com/sdroege/gstreamer-rs/pull/82 allowed using concrete types for queries directly, with this impact on attribute declaration:

Before:
``` rust
    pub struct MyContext {
        [...]
        position_query: gst::Query,
    }
```

After:
``` rust
    pub struct MyContext {
        [...]
        position_query: gst::query::Position<gst::Query>,
    }
```

This commit proposes to simplify the declaration to this:
``` rust
    pub struct MyContext {
        [...]
        position_query: gst::query::Position,
    }
```

This way, the user doesn't have to be aware of internal details.

### DIFF
--- a/gstreamer/src/query.rs
+++ b/gstreamer/src/query.rs
@@ -33,46 +33,46 @@ unsafe impl MiniObject for QueryRef {
 }
 
 impl GstRc<QueryRef> {
-    pub fn new_position(fmt: ::Format) -> Position<Self> {
+    pub fn new_position(fmt: ::Format) -> PositionImpl<Self> {
         assert_initialized_main_thread!();
         unsafe {
-            Position::<Self>(
+            PositionImpl::<Self>(
                 from_glib_full(ffi::gst_query_new_position(fmt.to_glib()))
             )
         }
     }
 
-    pub fn new_duration(fmt: ::Format) -> Duration<Self> {
+    pub fn new_duration(fmt: ::Format) -> DurationImpl<Self> {
         assert_initialized_main_thread!();
         unsafe {
-            Duration::<Self>(
+            DurationImpl::<Self>(
                 from_glib_full(ffi::gst_query_new_duration(fmt.to_glib()))
             )
         }
     }
 
-    pub fn new_latency() -> Latency<Self> {
+    pub fn new_latency() -> LatencyImpl<Self> {
         assert_initialized_main_thread!();
         unsafe {
-            Latency::<Self>(
+            LatencyImpl::<Self>(
                 from_glib_full(ffi::gst_query_new_latency())
             )
         }
     }
 
-    pub fn new_seeking(fmt: ::Format) -> Seeking<Self> {
+    pub fn new_seeking(fmt: ::Format) -> SeekingImpl<Self> {
         assert_initialized_main_thread!();
         unsafe {
-            Seeking::<Self>(
+            SeekingImpl::<Self>(
                 from_glib_full(ffi::gst_query_new_seeking(fmt.to_glib()))
             )
         }
     }
 
-    pub fn new_segment(fmt: ::Format) -> Segment<Self> {
+    pub fn new_segment(fmt: ::Format) -> SegmentImpl<Self> {
         assert_initialized_main_thread!();
         unsafe {
-            Segment::<Self>(
+            SegmentImpl::<Self>(
                 from_glib_full(ffi::gst_query_new_segment(fmt.to_glib()))
             )
         }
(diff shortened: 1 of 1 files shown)


sdroege:
Do you have an example where this would actually happen that you want to store a concrete query implementation in some struct?

fengalin:
I did this in [my application](https://github.com/fengalin/media-toc) for reuse purposes, in order to avoid re-instantiating the position query every time the position is refreshed. I implemented this after reading the [documentation for `gst_element_query_position`](https://developer.gnome.org/gstreamer/stable/GstElement.html#gst-element-query-position):

> If one repeatedly calls this function one can also create a query and reuse it in `gst_element_query()`.

I agree this is a cheap optimization and could live without it. Do you think this is wrong?

sdroege:
I've never done that so far and didn't even think of that :) For position and some other queries that will work fine, other queries will probably cause problems if reused (e.g. allocation query has an array of fields, and nothing would ever reset that to an empty array when re-using a query).

Maybe we could create `type PositionOwned = Position<Query>`, `type PositionRef = Position<&QueryRef>`, `type PositionRefMut = Position<&mut QueryRef>` but it seems not too useful to me.

fengalin:
Ok, let's not do it then.

sdroege:
ok :) Are there any other things in your application that are a bit awkward API-wise?

fengalin:
One thing that I have in mind is something similar to what we did for queries, but for TOCs & Tags. For instance, today we have to go through this:
``` rust
    let mut tags = TagList::new();
    {
        let tags = tags.get_mut().unwrap();
        tags.add::<Title>(&"some title", TagMergeMode::Append);
    }
```
I was thinking about implementing the same approach as in https://github.com/sdroege/gstreamer-rs/pull/82 so that we could avoid calling `get_mut()` for owned Tags/Tocs. What do you think? Maybe that would also be applicable to other structs (`Buffer`, `Sample`?) but I didn't run into other situations yet.

sdroege:
Would that simplify things a lot? And we have to make sure than that everything that takes a new reference of the query/buffer/sample/etc would consume the owned one, or requires the refcounted one instead (to prevent having a mutable reference around while someone else also already owns another reference). With queries that's relatively simple, other things need some more care :)

fengalin:
That would be more about least astonishment than about simplicity. On a Rust perspective, I think it seems weird declaring a `mut` instance, then needing to `get_mut` to modify it. Maybe using the original `GstRc<TagList>` would be less surprising.

I know you referred to concerns about the `ref_count` several times before and I have to admit my experience with GLib/GStreamer is still young. I'm trying to figure out use cases where the borrow checker wouldn't be sufficient to make sure the `ref_count` is not respected. Are you thinking about cases where the miniobject would be passed to a `C` function and that function would keep a reference on it?

sdroege:
> That would be more about least astonishment than about simplicity. On a Rust perspective, I think it seems weird declaring a mut instance, then needing to get_mut to modify it. Maybe using the original GstRc<TagList> would be less surprising.

It's not different from `let mut thing = Arc::new(Thing::new())`

> Are you thinking about cases where the miniobject would be passed to a C function and that function would keep a reference on it?

Basically all cases where in "normal" Rust you need to use Arc/Rc instead of having the borrow checker handle things for you. All these things can be referenced from multiple places at once, e.g. you (an element) could send a taglist downstream in an event and keep a reference to it. The event has another reference. The next elements might keep yet more references to the taglist. Only as long as someone has the only reference, the taglist is writable. Otherwise a copy has to be made first.

It's the more dynamic, general case of the borrow checker. You can have multiple read-only references at a time, but only ever a single mutable. But the borrow checker can't really know the situation at any point in time, it has to be checked at runtime (or you would have to create copies all the time).

There was also a very nice article from the Servo team about how using Arc like that improved performance for CSS layouting a lot. [This](https://blog.rust-lang.org/2017/11/14/Fearless-Concurrency-In-Firefox-Quantum.html) one IIRC.

> Maybe using the original GstRc<TagList> would be less surprising.

You mean always using the `GstRc<TagListRef>` type instead of the type alias, because it contains the information that it's actually reference counted?

> Are you thinking about cases where the miniobject would be passed to a C function and that function would keep a reference on it?

It does not necessary have to be a C function, the same can happen in pure Rust too. You have shared ownership of something, but might want to write to it and only ever create a copy if you have to. That is, if others are still keeping a reference to your thing.

fengalin:
Thanks, that's clear to me now :)

>  You mean always using the GstRc<TagListRef> type instead of the type alias, because it contains the information that it's actually reference counted?

Yes, that's what I meant.

PR #67, opened 2017-12-24
Wraps pbutils library

thiagoss:
Adds gstreamer-pbutils crate wrapping most of the features. A few caps parsing related functions are still not available as it needs some more gobject-introspection improvements.

PR #84, opened 2018-02-20
pbutils: add discoverer APIs

thiagoss:
First part of pbutils wrap, containing discoverer.

Done in pieces to make it easier to do proper review of the APIs.


thiagoss:
I'll be bringing the comments from the previous PR here and replying to keep the context of the initial discussion for those parts. (https://github.com/sdroege/gstreamer-rs/pull/67)

gstreamer-pbutils/src/auto/discoverer.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+unsafe impl Send for Discoverer {}
+unsafe impl Sync for Discoverer {}
+
+pub trait DiscovererExt {
+    fn discover_uri(&self, uri: &str) -> Result<DiscovererInfo, Error>;
+
+    fn discover_uri_async(&self, uri: &str) -> Result<(), glib::error::BoolError>;
```
thiagoss: This is never going to return an Error but I'd prefer to keep it as is to be in sync with the C API.

gstreamer-pbutils/src/auto/discoverer.rs
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    fn stop(&self) {
+        unsafe {
+            ffi::gst_discoverer_stop(self.to_glib_none().0);
+        }
+    }
+
+    fn connect_discovered<F: Fn(&Self, &DiscovererInfo, &Option<Error>) + Send + Sync + 'static>(&self, f: F) -> SignalHandlerId {
```
thiagoss: DiscovererInfo is always created, so it is returned even if with no information at all.

gstreamer-pbutils/src/auto/discoverer_info.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    fn get_streams(&self, streamtype: glib::types::Type) -> Vec<DiscovererStreamInfo>;
+
+    fn get_subtitle_streams(&self) -> Vec<DiscovererStreamInfo>;
+
+    fn get_tags(&self) -> Option<gst::TagList>;
+
+    fn get_toc(&self) -> Option<gst::Toc>;
+
+    fn get_uri(&self) -> Option<String>;
```
thiagoss: It seems like we don't have guards for NULL URIs on the input and it will lead to an Error. So this get needs to deal with that. The other option is to add checks at the API for rejecting NULLs, if that can be considered a non-breaking change.

gstreamer-pbutils/src/auto/discoverer.rs
sdroege: Ok

sdroege:
Thanks, that all makes sense. I'll take a closer look at the remaining changes later today but it's probably all good now (also I'll fix the build failure in a separate commit unless you're faster).

Gir_GstPbutils.toml
```diff
@@ -0,0 +1,137 @@
...
+
+manual = [
+    "GLib.Error",
+    "GLib.Source",
+    "GLib.DateTime",
+    "GLib.Variant",
+    "GObject.Object",
+    "Gst.Object",
+    "Gst.Element",
+    "Gst.Toc",
```
sdroege: This probably also needs the same as caps

gstreamer-pbutils/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "gstreamer-pbutils"
+version = "0.11.0"
+authors = ["Sebastian Dröge <sebastian@centricular.com>"]
```
sdroege: Add yourself too here if you want :)

gstreamer-pbutils/build.rs
```diff
@@ -0,0 +1,34 @@
...
+    stripper_lib::loop_over_files(
+        path.as_ref(),
+        &mut |w, s| stripper_lib::strip_comments(w, s, &mut io::sink(), true),
+        &ignores,
+        false,
+    );
+
+    #[cfg(feature = "embed-lgpl-docs")]
+    {
+        let docs = include_str!("../docs/gstreamer-pbutils/docs.md");
```
sdroege: This is why CI fails, you need to create this file with gir running in "-m doc" mode

gstreamer-pbutils/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+
+macro_rules! callback_guard {
+    () => (
+        let _guard = ::glib::CallbackGuard::new();
+    )
+}
+
+macro_rules! assert_initialized_main_thread {
+    () => (
+        if unsafe {::gst_ffi::gst_is_initialized()} != ::glib_ffi::GTRUE {
```
sdroege: We should call `gst_pb_utils_init` here also, but there's nothing to check if it was already called before...

sdroege: OTOH it does not actually do much so we might as well just always call it here

gstreamer-pbutils/src/auto/discoverer.rs
```diff
@@ -0,0 +1,149 @@
...
+            let ret = ffi::gst_discoverer_new(timeout.to_glib(), &mut error);
+            if error.is_null() { Ok(from_glib_full(ret)) } else { Err(from_glib_full(error)) }
+        }
+    }
+}
+
+unsafe impl Send for Discoverer {}
+unsafe impl Sync for Discoverer {}
+
+pub trait DiscovererExt {
```
sdroege: I believe we don't need the trait here

gstreamer-pbutils/src/auto/discoverer_stream_info.rs
```diff
@@ -0,0 +1,91 @@
...
+unsafe impl Sync for DiscovererStreamInfo {}
+
+pub trait DiscovererStreamInfoExt {
+    fn get_caps(&self) -> Option<gst::Caps>;
+
+    fn get_misc(&self) -> Option<gst::Structure>;
+
+    fn get_next(&self) -> Option<DiscovererStreamInfo>;
+
+    fn get_previous(&self) -> Option<DiscovererStreamInfo>;
```
sdroege: This looks like some kind of iterator impl around this would be nice. But that's for later :)

gstreamer-pbutils/src/discoverer.rs
```diff
@@ -0,0 +1,65 @@
...
+use glib::object::Downcast;
+
+use ffi;
+use glib_ffi;
+use gobject_ffi;
+
+use std::mem::transmute;
+use std::boxed::Box as Box_;
+
+pub trait DiscovererExtManual {
```
sdroege: If you remove the `DiscovererExt` trait, this can also just be an `impl Discoverer`

sdroege:
> Additionally I didn't add the new gir file with the fixes from today to gstreamer master. It changed more then what I had done and I was not sure if I should add everything or just the bits I modified. Let me know if those should be included in this PR

Add the 1.12 gir file with all the changes applied that you need here. Once 1.14 is out, I'll merge those in.

gstreamer-pbutils/src/lib.rs
sdroege: Related to this: https://bugzilla.gnome.org/show_bug.cgi?id=793611

sdroege: IMHO we should just call it here every time via a https://doc.rust-lang.org/std/sync/struct.Once.html

thiagoss: It doesn't hurt but I thought the plan was to deprecate the `gst_pb_utils_init()` and just have one for the specific parts that need it. Anyway, added the Once there.

sdroege: Yes but then the pbutils bindings would only work with 1.14. At least for the next 2 years I'd like to depend on a lower version than 1.14 (currently 1.8).

gstreamer-pbutils/src/auto/discoverer_audio_info.rs
```diff
@@ -25,13 +25,6 @@ impl DiscovererAudioInfo {
         }
     }
 
-    #[cfg(any(feature = "v1_14", feature = "dox"))]
-    pub fn get_channel_mask(&self) -> u64 {
```
sdroege: Why did this disappear?

gstreamer-pbutils/src/auto/discoverer_info.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn copy(&self) -> DiscovererInfo;
+
+    fn get_audio_streams(&self) -> Vec<DiscovererStreamInfo>;
+
+    fn get_container_streams(&self) -> Vec<DiscovererStreamInfo>;
+
+    fn get_duration(&self) -> gst::ClockTime;
+
+    #[cfg(any(feature = "v1_14", feature = "dox"))]
+    fn get_live(&self) -> bool;
```
sdroege: 1.14 feature here, let's stay at 1.12 API-wise for now (but annotation fixes from 1.14 are fine).

sdroege:
Looks good to me now (except for the 1.14 API). Unless you have any changes you'd like to add on top, I'd squash it all together later and merge it then :) Thanks a lot for your work and patience

thiagoss:
This is a nice journey to keep me practicing Rust. Glad to be helpful.
I'm going to try and squeeze in the iterator in the next minutes.

sdroege:
Great :) The iterator (also check if any of the other iterator traits can be implemented here) should probably be a new struct that is returned from some kind of `siblings` (better name ideally) function, not directly implemented on the type itself.

sdroege:
Or maybe a next_iter and prev_iter function?

sdroege:
> warning: use of deprecated item 'glib::CallbackGuard': Rustc has this functionality built-in since 1.24.0

You probably also want to update your GIR version so that these are not generated anymore (and then remove the macro from src/lib.rs)

gstreamer-pbutils/src/discoverer_stream_info.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+impl Iterator for DiscovererStreamInfoIterator {
+    type Item = DiscovererStreamInfo;
+
+    fn next(&mut self) -> Option<DiscovererStreamInfo> {
+        let current = self.stream_info.take();
+        self.stream_info = match &current {
+            &Some(ref c) => c.get_next(),
+            &None => None
```
sdroege: Why are the & needed here?

thiagoss: because we want to return the `current` in the end. And it wouldn't be possible as it would have been used/moved in the match statement.

thiagoss:
1st attempt at an iterator. If this is the right way I'll do the reverse one as well.

Also, this iteration starts on the next element, how do we iterate on all? I guess we also want an iterator on the `DiscovererInfo` level. The current one will make users have to get the first `DiscovererStreamInfo`, use it and then do the rest inside of the iterator processing.

gstreamer-pbutils/src/discoverer_stream_info.rs
```diff
@@ -0,0 +1,36 @@
...
+        };
+        current
+    }
+}
+
+impl DiscovererStreamInfo {
+    pub fn next_iter(&self) -> DiscovererStreamInfoIterator {
```
sdroege: And one for prev maybe. You can just store a boolean for the direction in the iterator, and it will then be optimized away by the compiler.

Also you could call this just Iterator or Iter, and not re-export it from src/lib.rs

sdroege:
> Also, this iteration starts on the next element, how do we iterate on all? I guess we also want an iterator on the DiscovererInfo level. The current one will make users have to get the first DiscovererStreamInfo, use it and then do the rest inside of the iterator processing.

I don't actually know what this API returns. It talks about a "chain". Will it give you e.g. "ID3" -> "MP3", or ...?

I think as it is now is fine, just needs the reverse one. We can think of nicer things later, but this would be according to the C API right now.

gstreamer-pbutils/src/discoverer_stream_info.rs
thiagoss: Not sure I understood this `Also you could call this just Iterator or Iter, and not re-export it from src/lib.rs`

I'll re-read on the prelude and exporting to see how this works and understand why we can do that.

sdroege: You would then have `gst_pbutils::DiscovererStreamInfo` and `gst_pbutils::discoverer_stream_info::Iter`. So that would not conflict with anything then, and as users generally don't have to name the type but only get it from the function, there would not be any inconvenience.

Check the caps iterator in GStreamer for example.

PR #85, opened 2018-02-20
Toc: implement Debug trait

fengalin:


gstreamer/src/toc.rs
```diff
@@ -99,6 +100,15 @@ impl ToOwned for TocRef {
     }
 }
 
+impl fmt::Debug for TocRef {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("Toc")
+            .field("scope", &self.get_scope())
+            .field("tags", &self.get_tags())
```
sdroege: You probably also want to dump the entries here?

gstreamer/src/toc.rs
```diff
@@ -237,6 +247,20 @@ impl ToOwned for TocEntryRef {
...
+impl fmt::Debug for TocEntryRef {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("TocEntry")
+            .field("entry_type", &self.get_entry_type())
+            .field("uid", &self.get_uid())
+            .field("start_stop", &self.get_start_stop_times())
+            .field("tags", &self.get_tags())
+            .field("is_alternative", &self.is_alternative())
+            .field("is_sequence", &self.is_sequence())
+            .field("loop", &self.get_loop())
```
sdroege: Sub entries here?

fengalin: I don't know. In that case, this should be recursive. I though the user would choose how and when to traverse the tree.

sdroege: Debug impls usually recurse into all their children

fengalin: Ok then.

PR #92, opened 2018-03-16
Element: add feature gate for test_call_async

fengalin:


fengalin:
CI fails due to GStreamer version being 1.12.2.

sdroege:
Oops!

sdroege:
Merged it manually, into master though.

sdroege:
> CI fails due to GStreamer version being 1.12.2.

I'm going to update once 1.14.0 is out soon :) But even then it will still fail due to the pollfd API in GstBus: --all-features would enable the Windows and Unix version, and then the build fails because the same function is defined twice.

I think I'll have to move away from --all-features for travis somehow. Do you have any suggestions?

fengalin:
I think I do :). See https://github.com/sdroege/gstreamer-rs/pull/94

PR #94, opened 2018-03-16
Bus: fix feature gate unix / windows

fengalin:
OS dependent conditions were tied to `dox` which would bypass them when using feature `dox` or `--all-features`


fengalin:
Or is it an issue because we will get the doc for one variant only?

sdroege:
Yes but I don't see a better way. Thanks!

fengalin:
Wait, maybe there is something to do similar to [this](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/mod.rs)
I'll try it tomorrow

PR #96, opened 2018-03-18
pbutils: adding encoding-profile bindings

thiagoss:


thiagoss:
This requires a small patch to pbutils in the lib, will push it after the release freeze is over

gstreamer-pbutils/src/auto/encoding_container_profile.rs
```diff
@@ -0,0 +1,68 @@
...
+            from_glib_full(ffi::gst_encoding_container_profile_new(name.0, description.0, format.to_glib_none().0, preset.0))
+        }
+    }
+}
+
+unsafe impl Send for EncodingContainerProfile {}
+unsafe impl Sync for EncodingContainerProfile {}
+
+pub trait EncodingContainerProfileExt {
+    fn add_profile<P: IsA<EncodingProfile>>(&self, profile: &P) -> Result<(), glib::error::BoolError>;
```
sdroege: This one has API to modify it, but no mutexes seem to be used. So this can't be Sync, and because it's refcounted also not Send.

I would suggest to make all these objects immutable and move the mutable parts into a builder for them, so that they can be built up in the beginning but from that point onwards are immutable

gstreamer-pbutils/src/auto/encoding_profile.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    fn get_preset_name(&self) -> Option<String>;
+
+    fn get_restriction(&self) -> Option<gst::Caps>;
+
+    fn get_type_nick(&self) -> Option<String>;
+
+    fn is_enabled(&self) -> bool;
+
+    fn set_allow_dynamic_output(&self, allow_dynamic_output: bool);
```
sdroege: Same thing here

Gir_GstPbutils.toml
```diff
@@ -140,3 +140,66 @@ trait = false
 name = "GstPbutils.DiscovererContainerInfo"
 status = "generate"
 trait = false
+
+[[object]]
+name = "GstPbutils.EncodingProfile"
+status = "generate"
+    [[object.function]]
+    name = "is_equal"
+    ignore = true
```
sdroege: Why? Please add at least a comment :)

sdroege:
Thanks, generally looks good except for the mutability/thread-safety mess-up that is inherited from the C API :)

thiagoss:
Tests are failing because of a problem on refcounting of the `restriction` caps field. It is marked as transfer-full (and allow none, which seems to be a bug as there is a GST_IS_CAPS check in the C source).

I noticed both autogenerated codes for `set_format` and `set_restriction` handle the caps as a reference and format is transfer-none and restriction is transfer-full. Shouldn't they have different handling of input parameters? The problem is that the restriction caps reference is not being lost after it is transfered to the encoding profile and later rust will try to unref it.

gstreamer-pbutils/src/encoding_profile.rs
```diff
@@ -0,0 +1,574 @@
...
+        unsafe {
+            ffi::gst_encoding_profile_set_preset_name(self.to_glib_none().0, preset_name.0);
+        }
+    }
+
+    fn set_restriction<'a, P: Into<Option<&'a gst::Caps>>>(&self, restriction: P) {
+        let restriction = restriction.into();
+        let restriction = restriction.to_glib_none();
+        unsafe {
+            ffi::gst_encoding_profile_set_restriction(self.to_glib_none().0, restriction.0);
```
sdroege: As you say, this is `transfer full`. The solution here would be to use `restriction.to_glib_full()`, which is giving a new reference

sdroege: And the C code is wrong and should do
```c
g_return_if_fail (restriction == NULL || GST_IS_CAPS (restriction));
```
so that it actually allows `NULL` caps. Maybe as a workaround we can convert `None` here to `gst::Caps::new_any()` (and fix it on the C side).

sdroege: Basically this here
```diff
diff --git a/gstreamer-pbutils/src/encoding_profile.rs b/gstreamer-pbutils/src/encoding_profile.rs
index 1ad7d4b..d3768a1 100644
--- a/gstreamer-pbutils/src/encoding_profile.rs
+++ b/gstreamer-pbutils/src/encoding_profile.rs
@@ -2,6 +2,7 @@
 use gst;
 use glib;
 use ffi;
+use gst_ffi;
 
 use std::error;
 use std::fmt;
@@ -96,9 +97,13 @@ impl<O: IsA<EncodingProfile> + IsA<glib::object::Object>> EncodingProfileEdit fo
 
     fn set_restriction<'a, P: Into<Option<&'a gst::Caps>>>(&self, restriction: P) {
         let restriction = restriction.into();
-        let restriction = restriction.to_glib_none();
         unsafe {
-            ffi::gst_encoding_profile_set_restriction(self.to_glib_none().0, restriction.0);
+            let restriction = match restriction {
+                Some(restriction) => restriction.to_glib_full(),
+                None => gst_ffi::gst_caps_new_any(),
+            };
+
+            ffi::gst_encoding_profile_set_restriction(self.to_glib_none().0, restriction);
         }
     }
 }
```
Fixes the warning at least

thiagoss: Changed the code for this. Thanks for the diff. Will push a patch for the C code as well.

thiagoss:
Thanks for the reminder, forgot to set my default email in github.

Gir_GstPbutils.toml
sdroege: @thiagoss ? :)

Gir_GstPbutils.toml
```diff
@@ -141,3 +141,93 @@ trait = false
...
+
+    # Ignore all setters, making it imutable. A builder will be provided.
```
sdroege: immutable has two 'm'

Gir_GstPbutils.toml
```diff
@@ -141,3 +141,93 @@ trait = false
...
+    ignore = true
+
+[[object]]
+name = "GstPbutils.EncodingTarget"
+status = "generate"
+
+    [[object.function]]
+    name = "add_profile"
+        [object.function.return]
+        bool_return_is_error = "Failed to add profile"
```
sdroege: Should this maybe also be a builder or is it acceptable to add profiles to the target at any time?

examples/src/bin/encodebin.rs
```diff
@@ -0,0 +1,271 @@
...
+        .build()?;
+
+    let container_profile = pbutils::EncodingContainerProfileBuilder::new()
+        .name("container")
+        .format(&gst::Caps::new_simple(
+            "video/x-matroska",
+            &[],
+        ))
+        .add_profile(&(video_profile.upcast()))
+        .add_profile(&(audio_profile.upcast()))
```
sdroege: This should probably be
```rust
fn add_profile<O: IsA<EncodingProfile>>(self, profile: &T) -> Self
```

Then you don't have to cast yourself here

examples/src/bin/encodebin.rs
```diff
@@ -0,0 +1,271 @@
...
+        .build()?;
+
+    encodebin.set_property("profile", &container_profile)?;
```
sdroege: This should be an `expect` or `unwrap`. It only can possibly fail here if you did a programming mistake

examples/src/bin/encodebin.rs
```diff
@@ -0,0 +1,271 @@
...
+        std::process::exit(-1)
+    };
+
+    let pipeline = gst::Pipeline::new(None);
+    let src = gst::ElementFactory::make("uridecodebin", None).ok_or(MissingElement("uridecodebin"))?;
+    let encodebin = gst::ElementFactory::make("encodebin", None).ok_or(MissingElement("encodebin"))?;
+    let sink = gst::ElementFactory::make("filesink", None).ok_or(MissingElement("filesink"))?;
+
+    src.set_property("uri", &uri)?;
+    sink.set_property("location", &output_file)?;
```
sdroege: Same for these

examples/src/bin/encodebin.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    configure_encodebin(&encodebin)?;
+
+    pipeline.add_many(&[&src, &encodebin, &sink])?;
```
sdroege: And this, in general

examples/src/bin/encodebin.rs
```diff
@@ -0,0 +1,271 @@
...
+                        ("Failed to get media type from pad {}", dbin_src_pad.get_name())
+                    );
+
+                    return;
+                }
+                Some(media_type) => media_type,
+            }
+        };
+
+        let insert_sink = |is_audio, is_video| -> Result<(), Error> {
```
sdroege: Not really "inserting a sink" but `link_to_encodebin` maybe?

gir-files/GstPbutils-1.0.gir
```diff
@@ -1684,8 +1684,11 @@ information like name, description, format and preset.</doc>
             <doc xml:space="preserve">The name of the target</doc>
             <type name="utf8" c:type="const gchar*"/>
           </parameter>
-          <parameter name="profilename" transfer-ownership="none">
-            <doc xml:space="preserve">(allow-none): The name of the profile, if %NULL
+          <parameter name="profilename"
+                     transfer-ownership="none"
+                     nullable="1"
+                     allow-none="1">
+            <doc xml:space="preserve">The name of the profile, if %NULL
```
sdroege: Please also fix on the C side

gstreamer-pbutils/src/auto/encoding_profile.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    fn get_type_nick(&self) -> Option<String>;
+
+    fn is_enabled(&self) -> bool;
+
+    fn get_property_restriction_caps(&self) -> Option<gst::Caps>;
+
+    fn set_property_restriction_caps(&self, restriction_caps: Option<&gst::Caps>);
+
+    fn connect_property_restriction_caps_notify<F: Fn(&Self) + Send + Sync + 'static>(&self, f: F) -> SignalHandlerId;
```
sdroege: This notify shouldn't exist, it can't change later. See https://github.com/gtk-rs/gir/pull/641 for how to disable generation of it

gstreamer-pbutils/src/auto/encoding_profile.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    fn get_description(&self) -> Option<String>;
+
+    fn get_file_extension(&self) -> Option<String>;
+
+    fn get_format(&self) -> gst::Caps;
+
+    fn get_input_caps(&self) -> gst::Caps;
+
+    fn get_name(&self) -> Option<String>;
```
sdroege: Can this really be `None`, and the description?

gstreamer-pbutils/src/encoding_profile.rs
```diff
@@ -0,0 +1,578 @@
...
+use glib::Cast;
+use glib::object::IsA;
+use glib::translate::*;
+
+use auto::EncodingProfile;
+use auto::EncodingAudioProfile;
+use auto::EncodingVideoProfile;
+use auto::EncodingContainerProfile;
+
+trait EncodingProfileEdit {
```
sdroege: `EncodingProfileBuilderCommon` maybe?

gstreamer-pbutils/src/encoding_profile.rs
```diff
@@ -0,0 +1,578 @@
...
+    fn description(&self) -> &str {
+        "invalid parameters to build encoding profile"
+    }
+
+    fn cause(&self) -> Option<&error::Error> {
+        None
+    }
+}
+
+struct EncodingProfileBuilderData<'a> {
```
sdroege: `...BuilderCommonData` maybe? Or `..BuilderCommon`?

gstreamer-pbutils/src/encoding_profile.rs
```diff
@@ -0,0 +1,578 @@
...
+    profile.set_preset(base_data.preset);
+    profile.set_preset_name(base_data.preset_name);
+    profile.set_allow_dynamic_output(base_data.allow_dynamic_output);
+    profile.set_enabled(base_data.enabled);
+    profile.set_restriction(base_data.restriction);
+    profile.set_presence(base_data.presence);
+}
+
+pub struct EncodingAudioProfileBuilder<'a> {
+    base : EncodingProfileBuilderData<'a>
```
sdroege: You might want to run rustfmt :)

gstreamer-pbutils/src/encoding_profile.rs
```diff
@@ -0,0 +1,578 @@
...
+        self.pass = pass;
+        self
+    }
+
+    pub fn variable_framerate(mut self, variable_framerate: bool) -> Self {
+        self.variable_framerate = variable_framerate;
+        self
+    }
+
+    pub fn build(self) -> Result<EncodingProfile, EncodingProfileBuilderError> {
```
sdroege: It would be nicer if this returned a `Result<EncodingVideoProfile, _>`

gstreamer-pbutils/src/encoding_profile.rs
```diff
@@ -0,0 +1,578 @@
...
+        }
+
+        let video_profile = EncodingVideoProfile::new(
+            self.base.format.unwrap(), self.base.preset, self.base.restriction, self.base.presence);
+
+        video_profile.set_pass(self.pass);
+        video_profile.set_variableframerate(self.variable_framerate);
+
+        let mut profile = video_profile.upcast();
+        set_common_fields(&mut profile, &self.base);
```
sdroege: You could use `upcast_ref()` here probably, or like I mentioned earlier with a generic function over the `IsA` trait

Why does this need a mutable reference here, encoding profiles conceptually do interior mutability so an immutable reference should be fine :)

gstreamer-pbutils/src/encoding_profile.rs
```diff
@@ -0,0 +1,578 @@
...
+        Ok(profile)
+    }
+}
+
+pub struct EncodingContainerProfileBuilder<'a> {
+    base : EncodingProfileBuilderData<'a>,
+    profiles: LinkedList<& 'a EncodingProfile>
```
sdroege: Better use a `Vec<_>`, `LinkedList` is almost never the right choice of data structure

gstreamer-pbutils/src/encoding_profile.rs
```diff
@@ -0,0 +1,578 @@
...
+        for profile in self.profiles {
+            container_profile.add_profile(profile).or_else(|_error| Err(EncodingProfileBuilderError))?;
+        }
+
+        let mut profile = container_profile.upcast();
+        set_common_fields(&mut profile, &self.base);
+        Ok(profile)
+    }
+
+    pub fn add_profile(mut self, profile: & 'a EncodingProfile) -> Self {
```
sdroege: As mentioned above, make this generic over the `IsA` trait :) Makes life easier for users

Basically, in the end there should be no explicit casts anymore anywhere

gstreamer-pbutils/src/auto/encoding_profile.rs
thiagoss: do we want to keep the property at all if the object is immutable?

gstreamer-pbutils/src/auto/encoding_profile.rs
thiagoss: yes, both can be set to NULL currently on the C side.

gir-files/GstPbutils-1.0.gir
thiagoss: I didn't get what is wrong with this function (encoding_profile_find).

gstreamer-pbutils/src/auto/encoding_profile.rs
sdroege: I don't think so

gir-files/GstPbutils-1.0.gir
sdroege: I'll check

sdroege:
Looks good to me, I'll merge it now :)

sdroege:
And will remove the automatically inserted docs, and look at the broken function

gir-files/GstPbutils-1.0.gir
sdroege: What do you mean with it being wrong?

Also please don't forget to add annotation fixes to GStreamer as needed :)

PR #100, opened 2018-03-25
Add manual SDPmessage definition

notbdu:
- Removes auto generated `SDPMessage`
- Adds a manual `SDPMessage` definition


gstreamer-sdp/src/s_d_p_key.rs
```diff
@@ -0,0 +1,11 @@
...
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use ffi;
+
+pub struct SDPKey(ffi::GstSDPKey);
```
notbdu: I wasn't sure if just wrapping the ffi definition is fine here or not. There were no methods defined in the  -sys repo so I figured this was fine?

Also, I tried to auto gen these definition (and others) but I ran into the same error:
```
thread 'main' panicked at 'Missing memory management functions for GstSdp.SDPKey', src/codegen/record.rs:61:9
```
I'm guessing that's part of why these definitions had to be manually written?

Gir_GstSdp.toml
```diff
@@ -16,6 +16,7 @@ external_libraries = [
 ]
 
 generate = [
+    "GstSdp.MIKEYMessage",
```
notbdu: Not sure if it was okay to auto gen this definition.

gstreamer-sdp/src/s_d_p_key.rs
sdroege: Just wrapping the FFI definition is fine but you'll have to implement memory management correctly and also implement the relevant traits from glib-rs. I didn't check if you did that here yet, just saying :) I'll take a closer look later.

For the memory management error, yes that's one of the reasons why these have to be manually implemented.

.gitignore
```diff
@@ -1,3 +1,5 @@
 /target/
 **/*.rs.bk
 Cargo.lock
+# vim swap files
+*.swp
```
sdroege: Please add this in a separate commit, or even as a separate PR against master :)

gstreamer-sdp/src/s_d_p_zone.rs
```diff
@@ -0,0 +1,27 @@
...
+// except according to those terms.
+
+use auto::SDPResult;
+
+use ffi;
+
+pub struct SDPZone(ffi::GstSDPZone);
+
+impl SDPZone {
+    pub fn set(&mut self, adj_time: &str, typed_time: &str) -> SDPResult {
```
sdroege: This is part of the reason why this (and others can't be autogenerated). set/clear are actually constructor and destructor here, and the `SDPResult` return value should become some kind of `Result<_,_>` (I don't think `clear` can even fail). So there should be a `fn new(adj_time, typed_time) -> Result<Self, _>`, and `clear` just is called in `Drop::drop`.

Also there should be accessors for the fields of these structs

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -8,4 +39,389 @@ impl SDPMessage {
...
+		}
+	}
+
+	pub fn get_zone(&self, idx: u32) -> Option<SDPZone> {
+		unsafe {
+			from_glib_none(ffi::gst_sdp_message_get_zone())
+		}
+	}
+
+	pub fn init(&mut self) -> SDPResult {
```
sdroege: This one here is actually the constructor

sdroege: That is, you call `glib_ffi::g_malloc(mem::size_of::<GstSDPMessage>())`, then `init` and then return that

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -8,4 +39,389 @@ impl SDPMessage {
...
+
+    pub fn parse_buffer(data: &[u8], msg: &mut SDPMessage) -> SDPResult {
+        assert_initialized_main_thread!();
+        let size = data.len() as u32;
+        unsafe {
+            from_glib(ffi::gst_sdp_message_parse_buffer(data.to_glib_none().0, size, msg.to_glib_none_mut().0))
+        }
+    }
+
+    pub fn parse_uri(uri: &str, msg: &mut SDPMessage) -> SDPResult {
```
sdroege: This and `parse_buffer` are also constructors in disguise

sdroege:
Shortly looked over it, thanks a lot :) This goes in the wrong direction but there's still more to do

notbdu:
Thanks for the feedback @sdroege! I'm going to hack on it again later tonight for me (UTC -5:00).

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -1,11 +1,440 @@
...
         assert_initialized_main_thread!();
         unsafe {
             let msg = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMessage>());
-            let _ = ffi::gst_sdp_message_new(msg);
+            ffi::gst_sdp_message_new(msg);
+            from_glib_full(msg)
+        }
+    }
+
+	pub fn init() -> SDPMessage {
```
notbdu: So after I changed this to a constructor, I noticed that this is doing essentially the same thing as `new()`.

sdroege: Yeah, init() is unneeded in this case

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -1,11 +1,440 @@
...
+                SDPResult::ok => Some(from_glib_full(msg)),
+                _ => Err(result),
+            }
+		}
+    }
+
+    pub fn parse_uri(uri: &str, msg: &mut SDPMessage) -> Result<Self, SDPResult> {
+        assert_initialized_main_thread!();
+		unsafe {
+            let mut msg = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMessage>());
```
notbdu: So in some places I saw usage of `std::mem::uninitialized` so I wasn't sure if I'm supposed to use that or glib's `g_malloc` method. Although, my operating assumption was that the compiler would infer the memory size to allocate?

sdroege: malloc is needed if it should be heap allocated and there is a Drop impl that frees the data (in all cases when you use the glib_wrapper macro), otherwise you can simply stack allocate it

gstreamer-sdp/src/s_d_p_time.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    pub fn start(&self) -> &str {
+        CStr::from_ptr(self.0.start).to_str().unwrap()
+    }
+
+    pub fn stop(&self) -> &str {
+        CStr::from_ptr(self.0.stop).to_str().unwrap()
+    }
+
+    // TODO: How to convert glib::GArray into a string slice?
```
notbdu: I wasn't sure how to convert `glib::GArray` into a string slice. It seems like it _might_ be possible to use `std::slice::from_raw_parts`?

sdroege: Is it an actual string or what is it?
If it is a string, you can use CStr on the data pointer of the GArray.

notbdu:
Btw, I also wasn't sure what you meant by the following:
> Just wrapping the FFI definition is fine but you'll have to implement memory management correctly and also implement the relevant traits from glib-rs.

Mmm what memory management and traits from `glib-rs` are required by structs like `SDPOrigin` and `SDPKey`. It's hard to tell what additional traits/methods are required since both types belong to an `SDPMessage`.

gstreamer-sdp/src/s_d_p_time.rs
notbdu: It seems to be an array of `GChars`

sdroege:
> Mmm what memory management and traits from glib-rs are required by structs like SDPOrigin and SDPKey. It's hard to tell what additional traits/methods are required since both types belong to an SDPMessage.

I'll check when going through the commits, maybe nothing else is needed here after all.

gstreamer-sdp/src/s_d_p_time.rs
sdroege: It's an array of arbitrary data, the question is what GstSDPMessage is exactly putting in there. Which seem to be an array of strings, not a single string. So you could return a `Vec<&str>` here

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -1,11 +1,440 @@
...
+        }
+    }
+
+    pub fn insert_phone(&mut self, idx: i32, phone: &str) -> SDPResult {
+        unsafe {
+            from_glib(ffi::gst_sdp_message_insert_phone(self.to_glib_none_mut().0, idx, phone.to_glib_none().0))
+        }
+    }
+
+    pub fn insert_time(&mut self, idx: i32, time: &mut SDPTime) -> SDPResult {
```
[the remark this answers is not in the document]
notbdu: Ah! That would be a by product of blindly copying the auto generated code w/o actually thinking about it haha 😓. My bad, yea it doesn't make sense to make it mutable when I'm not actually manipulating it...

notbdu:
Thanks for the comments!

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -1,11 +1,440 @@
...
+        }
+    }
+
+    pub fn add_attribute<'a, P: Into<Option<&'a str>>>(&mut self, key: &str, value: P) -> SDPResult {
+        unsafe {
+            from_glib(ffi::gst_sdp_message_add_attribute(self.to_glib_none_mut().0, key.to_glib_none().0, value.into().to_glib_none().0))
+        }
+    }
+
+    pub fn add_bandwidth(&mut self, bwtype: &str, bandwidth: u32) -> SDPResult {
```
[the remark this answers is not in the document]
notbdu: Should I add a helper function somewhere to do that? Where do you think it should live?

gstreamer-sdp/src/s_d_p_time.rs
notbdu: How would I convert this https://docs.rs/glib-sys/0.5.0/glib_sys/struct.GArray.html into a `Vec<&str>`? It have the length there but I don't think I can do something like pointer arithmetic to iterate over all the data elements.

gstreamer-sdp/src/auto/s_d_p_media.rs
```diff
@@ -176,10 +178,14 @@ impl SDPMedia {
         }
     }
 
-    //#[cfg(any(feature = "v1_8_1", feature = "dox"))]
-    //pub fn parse_keymgmt(&self, mikey: /*Ignored*/MIKEYMessage) -> SDPResult {
-    //    unsafe { TODO: call ffi::gst_sdp_media_parse_keymgmt() }
-    //}
+    #[cfg(any(feature = "v1_8_1", feature = "dox"))]
+    pub fn parse_keymgmt(&self) -> (SDPResult, MIKEYMessage) {
```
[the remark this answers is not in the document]
notbdu: I'm going to move this out of the auto gen'ed files then.

gstreamer-sdp/src/auto/m_i_k_e_y_message.rs
```diff
@@ -0,0 +1,172 @@
...
+
+impl MIKEYMessage {
+    pub fn new() -> MIKEYMessage {
+        assert_initialized_main_thread!();
+        unsafe {
+            from_glib_full(ffi::gst_mikey_message_new())
+        }
+    }
+
+    //pub fn new_from_bytes(bytes: /*Ignored*/&glib::Bytes, info: /*Ignored*/&mut MIKEYDecryptInfo, error: /*Ignored*/Option<Error>) -> MIKEYMessage {
```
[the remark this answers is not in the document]
notbdu: Ah, so there aren't any memory management functions for `MIKEYDecryptInfo`, `MIKEYEncryptInfo` or `MIKEYMapSRTP`.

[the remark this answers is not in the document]
notbdu: Also, both decrypt/encrypt info are void pointers. Not sure what we would implement for those struct definitions.

gstreamer-sdp/src/s_d_p_message.rs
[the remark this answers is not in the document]
sdroege: It would just mean to do manual implementations of these functions and then doing `if res == ffi::GST_SDP_RESULT_OK { Ok(whatever) } else { Err(()) }`

gstreamer-sdp/src/s_d_p_time.rs
sdroege:
```rust
let arr = (*vec).data as *const *const c_char;
let vec = Vec::with_capacity((*arr).len as usize);
for i in 0..len {
    vec.push(CStr::new(arr.offset(i)));
}
```

or something like that

gstreamer-sdp/src/auto/m_i_k_e_y_message.rs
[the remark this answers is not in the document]
sdroege: As structs around the void pointers, `struct MIKEYDecryptInfo(*const c_void)` or similar

gstreamer-sdp/src/s_d_p_time.rs
notbdu: Ahhh didn't know about `offset`, thanks @sdroege !

gstreamer-sdp/src/auto/m_i_k_e_y_message.rs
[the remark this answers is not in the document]
notbdu: To avoid changes to `MIKEYMessage` getting wiped, I'm going to move it out of the auto-gen'ed files.

notbdu:
Btw, I can't compile it locally via `cargo build`. Getting the following error:
```
    Updating registry `https://github.com/rust-lang/crates.io-index`
    Updating git repository `https://github.com/gtk-rs/glib`
    Updating git repository `https://github.com/gtk-rs/sys`
    Updating git repository `https://github.com/sdroege/gstreamer-sys`
error: failed to load source for a dependency on `gstreamer-sys`

Caused by:
  Unable to update https://github.com/sdroege/gstreamer-sys?branch=1.14
```

notbdu:
So I realized that the mutable arguments in some of the `insert_*` functions were due to the ffi function requiring mutable arguments so I changed them back.

(e.g. `insert_attribute(&mut self, idx: i32, attr: &mut SDPAttribute)`)

gstreamer-sdp/src/m_i_k_e_y_message.rs
```diff
@@ -0,0 +1,206 @@
...
+        assert_initialized_main_thread!();
+        unsafe {
+            from_glib_full(ffi::gst_mikey_message_new())
+        }
+    }
+
+    pub fn new_from_bytes(bytes: &mut glib::Bytes, info: &mut MIKEYDecryptInfo, error: Option<Error>) -> MIKEYMessage {
+        unsafe {
+            let mut err = match error {
+                Some(err) => glib_ffi::g_error_new_literal(
```
notbdu: Wasn't sure how to convert a rust `Error` to a `glib::GError` so I ended up doing this... 😓

sdroege:
> Btw, I can't compile it locally via cargo build. Getting the following error

Rebase your branch against master, not the 1.14 or webrtc branch. Then it should be fine.

sdroege:
> So I realized that the mutable arguments in some of the insert_* functions were due to the ffi function requiring mutable arguments so I changed them back.

They don't just require mutable arguments, they take ownership of their arguments. So instead of a reference they should take the arguments by value. This seems to be the case for all insert and replace functions.

gstreamer-sdp/src/m_i_k_e_y_message.rs
sdroege: This should actually be `pub fn new_from_bytes(bytes: &mut glib::Bytes, info: &mut MIKEYDecryptInfo) -> Result<MIKEYMessage, Error>`, like in other functions with `glib::Error`.

See e.g. here https://github.com/sdroege/gstreamer-rs/blob/b6a80b59fa0c32f210267de9fcd1d3adf69f1d9e/gstreamer/src/auto/functions.rs#L155-L157

sdroege:
I'll go through the other changes later today, thanks :)

gstreamer-sdp/src/auto/m_i_k_e_y_message.rs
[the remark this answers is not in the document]
sdroege: You can keep the autogenerated file and just put the manual implementations of functions into a new file into a second `impl MIKEYMessage` block

gstreamer-sdp/src/auto/m_i_k_e_y_decrypt_info.rs
```diff
@@ -0,0 +1,11 @@
...
+// from gir-files (https://github.com/gtk-rs/gir-files @ ???)
+// DO NOT EDIT
+
+use ffi;
+use glib::translate::*;
+use glib_ffi;
+use std::mem;
+use std::ptr;
+
+pub struct MIKEYDecryptInfo(ffi::GstMIKEYDecryptInfo);
```
sdroege: This probably should get some field accessors at least, maybe a constructor. Not sure how it is actually used :)

gstreamer-sdp/src/auto/m_i_k_e_y_encrypt_info.rs
```diff
@@ -0,0 +1,11 @@
...
+// from gir-files (https://github.com/gtk-rs/gir-files @ ???)
+// DO NOT EDIT
+
+use ffi;
+use glib::translate::*;
+use glib_ffi;
+use std::mem;
+use std::ptr;
+
+pub struct MIKEYEncryptInfo(ffi::GstMIKEYEncryptInfo);
```
sdroege: Same

gstreamer-sdp/src/auto/m_i_k_e_y_map_s_r_t_p.rs
```diff
@@ -0,0 +1,9 @@
+// This file was generated by gir (https://github.com/gtk-rs/gir @ 7f5a2b5)
+// from gir-files (https://github.com/gtk-rs/gir-files @ ???)
+// DO NOT EDIT
+
+use ffi;
+use glib::translate::*;
+use glib_ffi;
+use std::mem;
+use std::ptr;
```
sdroege: This file is basically empty

gstreamer-sdp/src/auto/m_i_k_e_y_message.rs
```diff
@@ -53,9 +55,11 @@ impl MIKEYMessage {
         }
     }
 
-    //pub fn add_payload(&mut self, payload: /*Ignored*/&mut MIKEYPayload) -> bool {
-    //    unsafe { TODO: call ffi::gst_mikey_message_add_payload() }
-    //}
+    pub fn add_payload(&mut self, payload: &mut MIKEYPayload) -> bool {
```
sdroege: Why the mutable reference?

gstreamer-sdp/src/auto/m_i_k_e_y_payload.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl MIKEYPayload {
+    pub fn new(type_: MIKEYPayloadType) -> MIKEYPayload {
+        assert_initialized_main_thread!();
+        unsafe {
+            from_glib_full(ffi::gst_mikey_payload_new(type_.to_glib()))
+        }
+    }
+
+    pub fn kemac_add_sub(&mut self, newpay: &mut MIKEYPayload) -> bool {
```
sdroege: Why mutable reference?

gstreamer-sdp/src/auto/m_i_k_e_y_payload.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    pub fn sp_set(&mut self, policy: u32, proto: MIKEYSecProto) -> bool {
+        unsafe {
+            from_glib(ffi::gst_mikey_payload_sp_set(self.to_glib_none_mut().0, policy, proto.to_glib()))
+        }
+    }
+
+    //pub fn t_set(&mut self, type_: MIKEYTSType, ts_value: &[u8]) -> bool {
```
sdroege: Why is this not generated?

gstreamer-sdp/src/auto/m_i_k_e_y_payload.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    pub fn sp_get_n_params(&self) -> u32 {
+        unsafe {
+            ffi::gst_mikey_payload_sp_get_n_params(self.to_glib_none().0)
+        }
+    }
+
+    //pub fn sp_get_param(&self, idx: u32) -> /*Ignored*/Option<MIKEYPayloadSPParam> {
```
sdroege: This looks useful to have :)

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -223,15 +215,15 @@ impl SDPMessage {
         }
     }
 
-    pub fn insert_attribute(&mut self, idx: i32, attr: &mut SDPAttribute) -> SDPResult {
+    pub fn insert_attribute(&mut self, idx: i32, attr: &SDPAttribute) -> SDPResult {
```
sdroege: This is taking ownership of the argument from what I can see, so should be changed to do that. After passing to native code, no drop should be called so it should probably be just forgotten with `mem::forget`

gstreamer-sdp/src/m_i_k_e_y_message.rs
```diff
@@ -1,20 +1,22 @@
...
+    MIKEYPRFFunc,
+    MIKEYPayload,
+    MIKEYPayloadType,
+};
+use m_i_k_e_y_decrypt_info::MIKEYDecryptInfo;
+use m_i_k_e_y_encrypt_info::MIKEYEncryptInfo;
+use m_i_k_e_y_map_s_r_t_p::MIKEYMapSRTP;
 
 glib_wrapper! {
     pub struct MIKEYMessage(Boxed<ffi::GstMIKEYMessage>);
```
sdroege: Instead of moving the whole file, you can keep the autogenerated one and only have the additional manual things here

notbdu:
Hmm so It looks like I'll have to manually edit all of the `auto` files... 

I wanted to move the `auto` files out after manually editing them so they didn't include the `automatically generated...` comment in the head of the file and remove them from the `.toml`. Since all manual edits would get wiped if someone were to run gir again.

I can move them back if you think that makes sense tho!

sdroege:
Don't edit the automatically generated files. You can keep them as-is for the parts that are correct and just add an additional file with the manual code. See e.g. https://github.com/sdroege/gstreamer-rs/blob/master/gstreamer/src/auto/bus.rs and https://github.com/sdroege/gstreamer-rs/blob/master/gstreamer/src/bus.rs#L72

notbdu:
Oooohh okay! Sounds good, I'll do that then.

notbdu:
What about for methods like the following: 
```
pub fn insert_payload(&mut self, idx: u32, payload: &mut MIKEYPayload) -> bool {
    unsafe {
        from_glib(ffi::gst_mikey_message_insert_payload(self.to_glib_none_mut().0, idx, payload.to_glib_full()))
    }
}
```
This is part of `src/auto/m_i_k_e_y_message.rs`, but you mentioned that we want the signature to be `insert_payload(&mut self, idx: u32, payload: MIKEYPayload)` because we are consuming the payload variable and passing full ownership to gstreamer. Would I... comment out the auto-gen'ed function and write my own outside of the `auto` file?

notbdu:
Implementing the commented out functions sounds good to me tho! 👍

sdroege:
You can mark specific functions as ignored in the toml, see e.g. https://github.com/sdroege/gstreamer-rs/blob/master/Gir_Gst.toml#L132 . And then implement those correctly in the manual file.

notbdu:
Okay, I addressed your comments and also squashed my commits. This _should_ be close? 

I wasn't able to rebase off of master due to some nasty merge conflicts which would cause issues merging into your webrtc WIP branch as well.

notbdu:
I didn't put any of the commented out functions in the auto-gen'ed files into the `ignore` section of the gir toml. I just went ahead and manually implemented them outside of the `auto` dir.

sdroege:
> I didn't put any of the commented out functions in the auto-gen'ed files into the ignore section of the gir toml. I just went ahead and manually implemented them outside of the auto dir.

That's no problem. Just for the ones that are not commented out but need to be manually implemented you must ignore them explicitly.

sdroege:
I've rebased the `webrtc-sdp` branch against master, you should be able to rebase yours against that now and then it should be possible to build

gstreamer-sdp/src/auto/m_i_k_e_y_message.rs
```diff
@@ -0,0 +1,166 @@
...
+    //    unsafe { TODO: call ffi::gst_mikey_message_add_t() }
+    //}
+
+    pub fn add_t_now_ntp_utc(&mut self) -> bool {
+        unsafe {
+            from_glib(ffi::gst_mikey_message_add_t_now_ntp_utc(self.to_glib_none_mut().0))
+        }
+    }
+
+    pub fn base64_encode(&mut self) -> Option<String> {
```
sdroege: Does not have to be a mutable reference

gstreamer-sdp/src/auto/m_i_k_e_y_message.rs
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    pub fn find_payload(&self, type_: MIKEYPayloadType, nth: u32) -> Option<MIKEYPayload> {
+        unsafe {
+            from_glib_none(ffi::gst_mikey_message_find_payload(self.to_glib_none().0, type_.to_glib(), nth))
+        }
+    }
+
+    //pub fn get_cs_srtp(&self, idx: u32) -> /*Ignored*/Option<MIKEYMapSRTP> {
```
sdroege: There are still some ignored functions but let's worry about those once the basics are all merged

gstreamer-sdp/src/auto/m_i_k_e_y_message.rs
```diff
@@ -0,0 +1,166 @@
...
+        unsafe {
+            from_glib(ffi::gst_mikey_message_remove_payload(self.to_glib_none_mut().0, idx))
+        }
+    }
+
+    //pub fn replace_cs_srtp(&mut self, idx: i32, map: /*Ignored*/&MIKEYMapSRTP) -> bool {
+    //    unsafe { TODO: call ffi::gst_mikey_message_replace_cs_srtp() }
+    //}
+
+    pub fn set_info(&mut self, version: u8, type_: MIKEYType, V: bool, prf_func: MIKEYPRFFunc, CSB_id: u32, map_type: MIKEYMapType) -> bool {
```
sdroege: There are all kinds of setters in this type but no getters, I assume some field accessors would be needed. But that's also for a future PR

gstreamer-sdp/src/auto/m_i_k_e_y_payload.rs
sdroege: Also for a future PR

gstreamer-sdp/src/m_i_k_e_y_decrypt_info.rs
```diff
@@ -0,0 +1,23 @@
...
+use ffi;
+use glib::translate::*;
+use glib_ffi;
+use libc::c_void;
+
+pub struct MIKEYDecryptInfo(ffi::GstMIKEYDecryptInfo);
+
+impl MIKEYDecryptInfo {
+    pub fn new() -> MIKEYDecryptInfo {
+        MIKEYDecryptInfo(ffi::GstMIKEYDecryptInfo(ptr::null_mut() as *mut c_void))
```
sdroege: This and the other constructors that use NULL don't seem like a good idea. Just don't have a constructor, these would have to be created differently somehow... and there does not even seem to be a definition inside GStreamer for these structs, and they are not even used anywhere internally. You can safely make them in function arguments optional (`Option`, can also be configured in the toml).

gstreamer-sdp/src/m_i_k_e_y_message.rs
```diff
@@ -0,0 +1,98 @@
...
+    MIKEYMessage,
+    MIKEYTSType,
+};
+use m_i_k_e_y_payload::MIKEYPayload;
+use m_i_k_e_y_decrypt_info::MIKEYDecryptInfo;
+use m_i_k_e_y_encrypt_info::MIKEYEncryptInfo;
+use m_i_k_e_y_map_s_r_t_p::MIKEYMapSRTP;
+
+impl MIKEYMessage {
+    pub fn new_from_bytes(bytes: glib::Bytes, info: MIKEYDecryptInfo) -> Result<MIKEYMessage, Error> {
```
sdroege: Can take the bytes by reference. The function does not take ownership of it

gstreamer-sdp/src/m_i_k_e_y_payload.rs
```diff
@@ -0,0 +1,35 @@
...
+impl MIKEYPayload {
+
+    pub fn kemac_add_sub(&mut self, newpay: MIKEYPayload) -> bool {
+        unsafe {
+            from_glib(ffi::gst_mikey_payload_kemac_add_sub(self.to_glib_none_mut().0, newpay.to_glib_full()))
+        }
+        mem::forget(newpay);
+    }
+
+    //pub fn sp_get_param(&self, idx: u32) -> /*Ignored*/Option<MIKEYPayloadSPParam> {
```
sdroege: Why commented out functions in the manual file?

gstreamer-sdp/src/s_d_p_attribute.rs
```diff
@@ -0,0 +1,44 @@
...
+pub struct SDPAttribute(ffi::GstSDPAttribute);
+
+impl SDPAttribute {
+    pub fn new(key: &str, value: &str) -> Result<Self, SDPResult> {
+        assert_initialized_main_thread!();
+        unsafe {
+            let mut attr = mem::uninitialized();
+            let result = from_glib(ffi::gst_sdp_attribute_set(&mut attr, key.to_glib_none().0, value.to_glib_none().0));
+            match result {
+                SDPResult::Ok => Ok(SDPAttribute(attr)),
```
sdroege: As `SDPResult` is an enum with just `OK` and `INVAL`, you can make the return values of all these functions simply `Result<_, ()>`. Less noise, and we don't have to generate the `SDPResult` type at all then (only in the -sys bindings but there we have it already)

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -0,0 +1,448 @@
...
+    pub struct SDPMedia(Boxed<ffi::GstSDPMedia>);
+
+    match fn {
+        copy => |ptr| ffi::gst_sdp_media_copy(mut_override(ptr)),
+        free => |ptr| ffi::gst_sdp_media_free(ptr),
+    }
+}
+
+impl SDPMedia {
+    pub fn add_attribute<'a, P: Into<Option<&'a str>>>(&mut self, key: &str, value: P) -> Result<_, ()> {
```
sdroege: You need to provide a type for the successful case. This should be `Result<(), ()>` I guess

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -0,0 +1,448 @@
...
+            from_glib(ffi::gst_sdp_media_attributes_to_caps(self.to_glib_none().0, caps.to_glib_none().0))
+        };
+        match result {
+            SDPResult::ok => Some(_),
+            _ => Err(()),
+            }
+        }
+    }
+
+    pub fn bandwidths_len(&self) -> u32 {
```
sdroege: In a future PR it would make sense to get rid of this and the indexed getter (also for other fields), and return a FixedSizeIterator or a Vec

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -0,0 +1,448 @@
...
+        }
+    }
+
+    pub fn formats_len(&self) -> u32 {
+        unsafe {
+            ffi::gst_sdp_media_formats_len(self.to_glib_none().0)
+        }
+    }
+
+    pub fn get_attribute(&self, idx: u32) -> Option<SDPAttribute> {
```
sdroege: This does not give you ownership of the return value but borrows it to you. It should return a `Option<&SDPAttribute>`. Same for all the other getters

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -0,0 +1,448 @@
...
+        unsafe {
+            from_glib_none(ffi::gst_sdp_media_get_attribute(self.to_glib_none().0, idx) )
+        }
+    }
+
+    pub fn get_attribute_val(&self, key: &str) -> Option<String> {
```
sdroege: `Option<&str>`

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -0,0 +1,448 @@
...
+        }
+    }
+
+    pub fn get_bandwidth(&self, idx: u32) -> Option<SDPBandwidth> {
+        unsafe {
+            from_glib_none(ffi::gst_sdp_media_get_bandwidth(self.to_glib_none().0, idx))
+        }
+    }
+
+    pub fn get_caps_from_media(&self, pt: i32) -> Option<gst::Caps> {
```
sdroege: Only exception: the caps are indeed owned

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -0,0 +1,448 @@
...
+        }
+    }
+
+    pub fn get_proto(&self) -> Option<String> {
+        unsafe {
+            from_glib_none(ffi::gst_sdp_media_get_proto(self.to_glib_none().0))
+        }
+    }
+
+    pub fn init(&mut self) -> Result<_, ()> {
```
sdroege: This can go away, you have a constructor

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -0,0 +1,448 @@
...
+            from_glib(ffi::gst_sdp_media_init(self.to_glib_none_mut().0))
+        };
+        match result {
+            SDPResult::ok => Some(_),
+            _ => Err(()),
+            }
+        }
+    }
+
+    pub fn insert_attribute(&mut self, idx: i32, attr: &mut SDPAttribute) -> Result<_, ()> {
```
sdroege: This takes ownership of the argument, not just a mutable reference. And you should `mem::forget` it in the end

sdroege: Same for all the other insert/replace, except for the ones taking strings

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -0,0 +1,448 @@
...
+            from_glib(ffi::gst_sdp_media_set_proto(self.to_glib_none_mut().0, proto.to_glib_none().0))
+        };
+        match result {
+            SDPResult::ok => Some(_),
+            _ => Err(()),
+            }
+        }
+    }
+
+    pub fn uninit(&mut self) -> Result<_, ()> {
```
sdroege: Should go away, that's basically the drop impl

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -1,11 +1,617 @@
...
+}
 
 impl SDPMessage {
     pub fn new() -> SDPMessage {
         assert_initialized_main_thread!();
         unsafe {
             let msg = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMessage>());
-            let _ = ffi::gst_sdp_message_new(msg);
+            ffi::gst_sdp_message_new(msg);
             from_glib_full(msg)
```
sdroege: This function is actually wrong
```rust
let mut msg = mem::uninitialized();
ffi::gst_sdp_message_new(&mut msg);
from_glib_full(msg)
```

What you implemented here would be correct for calling `gst_sdp_message_init`, if you called `g_malloc0`

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -1,11 +1,617 @@
...
+            from_glib(ffi::gst_sdp_message_dump(self.to_glib_none().0))
+        };
+        match result {
+            SDPResult::ok => Some(_),
+            _ => Err(()),
+            }
+        }
+    }
+
+    pub fn emails_len(&self) -> u32 {
```
sdroege: Also these should become a FixedSizeIterator or at least Vec in a future PR

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -1,11 +1,617 @@
...
+        unsafe {
+            ffi::gst_sdp_message_emails_len(self.to_glib_none().0)
+        }
+    }
+
+    pub fn get_attribute(&self, idx: u32) -> Option<SDPAttribute> {
```
sdroege: Borrowing for all getters. You can do this by marking the struct as `#[repr(C)]` and then doing `&*(the_ptr as *mut SDPAttribute)`, i.e. casting from the FFI type to your wrapper struct

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -1,11 +1,617 @@
...
+            let mut msg = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMessage>());
+            let result = from_glib(ffi::gst_sdp_message_parse_buffer(data.to_glib_none().0, size, &mut msg));
+            match result {
+                SDPResult::ok => Some(from_glib_full(msg)),
+                _ => Err(result),
+            }
+        }
+    }
+
+    pub fn parse_uri(uri: &str, msg: &mut SDPMessage) -> Result<Self, SDPResult> {
```
sdroege: Should not take a `msg` parameter anymore

notbdu:
Thanks for the feedback!

notbdu:
Just a heads up, `cargo build` is still failing even after the rebase.

```
bdu @ Bos-MacBook-Pro.local gstreamer-sdp (bdu/webrtcsdp) $ cargo build --verbose
    Updating registry `https://github.com/rust-lang/crates.io-index`
    Updating git repository `https://github.com/gtk-rs/glib`
    Updating git repository `https://github.com/gtk-rs/sys`
    Updating git repository `https://github.com/sdroege/gstreamer-sys`
    Updating git repository `https://github.com/gtk-rs/gio`
    Updating git repository `https://github.com/sdroege/gstreamer-sys`
error: failed to load source for a dependency on `gstreamer-sys`

Caused by:
  Unable to update https://github.com/sdroege/gstreamer-sys?branch=1.14

Caused by:
  failed to find branch `1.14`

Caused by:
  cannot locate local branch '1.14'; class=Reference (4); code=NotFound (-3)
```

sdroege:
Even after cargo update? There should be no reference to the 1.14 branch in any Cargo.toml anymore.

I'll check later

notbdu:
Yea.. :/, when I run `cargo update` I see this:
```
bdu @ Bos-MacBook-Pro.local gstreamer-sdp (bdu/webrtcsdp) $ cargo update --verbose
    Updating registry `https://github.com/rust-lang/crates.io-index`
    Updating git repository `https://github.com/gtk-rs/glib`
    Updating git repository `https://github.com/gtk-rs/sys`
    Updating git repository `https://github.com/sdroege/gstreamer-sys`
    Updating git repository `https://github.com/gtk-rs/gio`
    Updating git repository `https://github.com/sdroege/gstreamer-sys`
error: failed to load source for a dependency on `gstreamer-sys`

Caused by:
  Unable to update https://github.com/sdroege/gstreamer-sys?branch=1.14

Caused by:
  failed to find branch `1.14`

Caused by:
  cannot locate local branch '1.14'; class=Reference (4); code=NotFound (-3)
```

sdroege:
In `gstreamer-webrtc/Cargo.toml`, change that from `branch = "1.14"` to master or complete change the branch parameter.

notbdu:
Oh sweet, I can see the > 1 million errors I have now haha. Will update PR w/ fixes when I'm finished.

gstreamer-sdp/src/m_i_k_e_y_message.rs
```diff
@@ -25,17 +25,17 @@ use m_i_k_e_y_encrypt_info::MIKEYEncryptInfo;
...
 
 impl MIKEYMessage {
-    pub fn new_from_bytes(bytes: glib::Bytes, info: MIKEYDecryptInfo) -> Result<MIKEYMessage, Error> {
+    pub fn new_from_bytes<'a, P: Into<Option<&'a MIKEYDecryptInfo>>>(bytes: &glib::Bytes, info: P) -> Result<MIKEYMessage, Error> {
         unsafe {
             let mut error = ptr::null_mut();
-            let ret = ffi::gst_mikey_message_new_from_bytes(bytes.to_glib_full(), info.to_glib_full(), &mut error);
+            let ret = ffi::gst_mikey_message_new_from_bytes(bytes.to_glib_none(), info.to_glib_full(), &mut error);
             mem::forget(bytes);
             mem::forget(info);
```
sdroege: This function does not take ownership of the bytes, same for the other one below with data.

The `info` is also not taken ownership of

gstreamer-sdp/src/m_i_k_e_y_payload.rs
```diff
@@ -25,11 +26,21 @@ impl MIKEYPayload {
         mem::forget(newpay);
     }
 
-    //pub fn sp_get_param(&self, idx: u32) -> /*Ignored*/Option<MIKEYPayloadSPParam> {
-    //    unsafe { TODO: call ffi::gst_mikey_payload_sp_get_param() }
-    //}
+    pub fn kemac_get_sub(&self, idx: u32) -> Option<MIKEYPayload> {
```
sdroege: This is missing a `&`

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -22,6 +22,7 @@ use s_d_p_connection::SDPConnection;
 use s_d_p_key::SDPKey;
 
 glib_wrapper! {
+    #[repr(C)]
```
sdroege: This does not need `#[repr(C)]`. For functions to return a reference to an `SDPMedia`, you would use `from_glib_borrow()`

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -31,47 +32,56 @@ glib_wrapper! {
 }
 
 impl SDPMedia {
-    pub fn add_attribute<'a, P: Into<Option<&'a str>>>(&mut self, key: &str, value: P) -> Result<_, ()> {
+    pub fn new() -> (SDPResult, SDPMedia) {
```
sdroege: `SDPResult` not really needed here, this should be `Result<Self, ()>`

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -118,27 +128,27 @@ impl SDPMedia {
...
+    pub fn get_attribute_val(&self, key: &str) -> Option<&str> {
         unsafe {
             from_glib_none(ffi::gst_sdp_media_get_attribute_val(self.to_glib_none().0, key.to_glib_none().0))
         }
     }
 
-    pub fn get_attribute_val_n(&self, key: &str, nth: u32) -> Option<String> {
+    pub fn get_attribute_val_n(&self, key: &str, nth: u32) -> Option<&str> {
         unsafe {
             from_glib_none(ffi::gst_sdp_media_get_attribute_val_n(self.to_glib_none().0, key.to_glib_none().0, nth))
```
sdroege: For getting `&str` references you need to use `CStr` directly

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -118,27 +128,27 @@ impl SDPMedia {
...
         }
     }
 
-    pub fn get_bandwidth(&self, idx: u32) -> Option<SDPBandwidth> {
+    pub fn get_bandwidth(&self, idx: u32) -> Option<&SDPBandwidth> {
         unsafe {
-            from_glib_none(ffi::gst_sdp_media_get_bandwidth(self.to_glib_none().0, idx))
+            &*(from_glib_none(ffi::gst_sdp_media_get_bandwidth(self.to_glib_none().0, idx)) as *mut SDPBandwidth)
```
sdroege: For all the other references, don't use `from_glib_none` either. This one here for example would be
```Rust
let ptr = ffi::gst_sdp_media_get_bandwidth(self.to_glib_none().0, idx);
if ptr.is_null() {
    None
} else {
    Some(&*(ptr as *mut SDPBandwidth))
}
```

(or maybe `as *const ...`)

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -190,249 +200,225 @@ impl SDPMedia {
...
-            _ => Err(()),
-            }
-        }
-    }
-
-    pub fn insert_attribute(&mut self, idx: i32, attr: &mut SDPAttribute) -> Result<_, ()> {
+    pub fn insert_attribute(&mut self, idx: i32, attr: SDPAttribute) -> Result<(), ()> {
         let result = unsafe {
-            from_glib(ffi::gst_sdp_media_insert_attribute(self.to_glib_none_mut().0, idx, attr.to_glib_none_mut().0))
+            from_glib(ffi::gst_sdp_media_insert_attribute(self.to_glib_none_mut().0, idx, attr.to_glib_full()))
```
sdroege: Not `to_glib_full` for all these as it would create a copy. Instead `&attr.0`

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -190,249 +200,225 @@ impl SDPMedia {
...
             _ => Err(()),
             }
         }
     }
 
-    pub fn replace_bandwidth(&mut self, idx: u32, bw: &mut SDPBandwidth) -> Result<_, ()> {
+    pub fn replace_bandwidth(&mut self, idx: u32, bw: SDPBandwidth) -> Result<(), ()> {
         let result = unsafe {
-            from_glib(ffi::gst_sdp_media_replace_bandwidth(self.to_glib_none_mut().0, idx, bw.to_glib_none_mut().0))
+            from_glib(ffi::gst_sdp_media_replace_bandwidth(self.to_glib_none_mut().0, idx, bw.to_glib_full()))
```
sdroege: Also for all these types you did not implement the GLib translation traits so this won't work. But you don't need the traits here, just some pointer casting for these types

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -35,85 +35,84 @@ impl SDPMessage {
     pub fn new() -> SDPMessage {
         assert_initialized_main_thread!();
         unsafe {
-            let msg = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMessage>());
-            ffi::gst_sdp_message_new(msg);
-            from_glib_full(msg)
+            let msg = mem::uninitialized();
+            from_glib_full(ffi::gst_sdp_message_new(&mut msg))
```
sdroege: This returns an `SDPResult`, not the actual message. Your message would be `msg`

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -165,167 +164,167 @@ impl SDPMessage {
         }
     }
 
-    pub fn get_attribute(&self, idx: u32) -> Option<SDPAttribute> {
+    pub fn get_attribute(&self, idx: u32) -> Option<&SDPAttribute> {
         unsafe {
-            from_glib_none(ffi::gst_sdp_message_get_attribute(self.to_glib_none().0, idx))
+            &*(from_glib_none(ffi::gst_sdp_message_get_attribute(self.to_glib_none().0, idx)) as *mut SDPAttribute)
```
sdroege: Just like above, store the pointer returned by the function, check for NULL and if so return None, otherwise the `&*` + cast

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -165,167 +164,167 @@ impl SDPMessage {
...
+    pub fn get_email(&self, idx: u32) -> Option<&str> {
         unsafe {
             from_glib_none(ffi::gst_sdp_message_get_email(self.to_glib_none().0, idx))
         }
     }
 
-    pub fn get_information(&self) -> Option<String> {
+    pub fn get_information(&self) -> Option<&str> {
         unsafe {
             from_glib_none(ffi::gst_sdp_message_get_information(self.to_glib_none().0))
```
sdroege: And for strings, check for NULL first and then use `CStr` directly

notbdu:
Thanks for the additional comments, they're very helpful! I'm still working through all the issues.... there's quite a few haha 😓 .

sdroege:
Don't worry and take your time, and ask if you have any questions :)

notbdu:
Alright, so I got it down to just 4 compilation errors now...
```
error[E0599]: no method named `to_glib_none_mut` found for type `&mut gst::CapsRef` in the current scope
   --> gstreamer-sdp/src/s_d_p_message.rs:127:91
    |
127 |             from_glib(ffi::gst_sdp_message_attributes_to_caps(self.to_glib_none().0, caps.to_glib_none_mut().0))
    |                                                                                           ^^^^^^^^^^^^^^^^

error[E0061]: this function takes 2 parameters but 1 parameter was supplied
  --> gstreamer-sdp/src/s_d_p_media.rs:31:23
   |
31 |         copy => |ptr| ffi::gst_sdp_media_copy(mut_override(ptr)),
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 2 parameters

error[E0308]: mismatched types
  --> gstreamer-sdp/src/s_d_p_media.rs:31:23
   |
27 | / glib_wrapper! {
28 | |     pub struct SDPMedia(Boxed<ffi::GstSDPMedia>);
29 | |
30 | |     match fn {
31 | |         copy => |ptr| ffi::gst_sdp_media_copy(mut_override(ptr)),
   | |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected *-ptr, found i32
32 | |         free => |ptr| ffi::gst_sdp_media_free(ptr),
33 | |     }
34 | | }
   | |_- expected `*mut ffi::GstSDPMedia` because of return type
   |
   = note: expected type `*mut ffi::GstSDPMedia`
              found type `i32`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

error[E0308]: mismatched types
  --> gstreamer-sdp/src/s_d_p_media.rs:32:23
   |
27 | / glib_wrapper! {
28 | |     pub struct SDPMedia(Boxed<ffi::GstSDPMedia>);
29 | |
30 | |     match fn {
31 | |         copy => |ptr| ffi::gst_sdp_media_copy(mut_override(ptr)),
32 | |         free => |ptr| ffi::gst_sdp_media_free(ptr),
   | |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^- help: try adding a semicolon: `;`
   | |                       |
   | |                       expected (), found i32
33 | |     }
34 | | }
   | |_- expected `()` because of default return type
   |
   = note: expected type `()`
              found type `i32`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

error: aborting due to 4 previous errors
```

notbdu:
A few questions:
- Should I add the `ToGlibPtr` trait impl to `CapsRef` or should I just expose the wrapped `ffi::GstCapsRef` struct?
- I ended up exposing all of the wrapped `ffi` structs to avoid having to impl all of the to glib traits... is that okay for now? I'm assuming we'd want to use the `glib_wrapper` macro on everything?
- It looks like the auto generated copy/free functions for `SDPMedia`are incorrect. I'm not sure what to do there...

sdroege:
```
error[E0599]: no method named `to_glib_none_mut` found for type `&mut gst::CapsRef` in the current scope
   --> gstreamer-sdp/src/s_d_p_message.rs:127:91
    |
127 |             from_glib(ffi::gst_sdp_message_attributes_to_caps(self.to_glib_none().0, caps.to_glib_none_mut().0))
    |     
```

Use `caps.as_mut_ptr()` here

```
error[E0308]: mismatched types
  --> gstreamer-sdp/src/s_d_p_media.rs:31:23
   |
27 | / glib_wrapper! {
28 | |     pub struct SDPMedia(Boxed<ffi::GstSDPMedia>);
29 | |
30 | |     match fn {
31 | |         copy => |ptr| ffi::gst_sdp_media_copy(mut_override(ptr)),
   | |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected *-ptr, found i32
32 | |         free => |ptr| ffi::gst_sdp_media_free(ptr),
33 | |     }
34 | | }
   | |_- expected `*mut ffi::GstSDPMedia` because of return type
   |
   = note: expected type `*mut ffi::GstSDPMedia`
              found type `i32`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

The signature of the functions is wrong. You could put a closure there that, that in the case of free ignores the return value and in case of copy first mallocs memory, then passes it there, and then returns it while ignoring the return value.

sdroege:
For your questions, I think all but this one are answered by my comment?

> I ended up exposing all of the wrapped ffi structs to avoid having to impl all of the to glib traits... is that okay for now? I'm assuming we'd want to use the glib_wrapper macro on everything?

We can't use `glib_wrapper` for most of these types and simply wrapping the ffi types is the only thing that can be done here. That's all ok :)

notbdu:
Okay, it compiles now w/ some warnings (mostly just unused code)...

sdroege:
Thanks a lot, especially for your patience. I'll go through everything tomorrow. Before merging, we should get rid of the warnings though but that should be easy :)

notbdu:
No problem, I should be thanking you for reviewing my PR and helping me work through the issues.

gstreamer-sdp/src/m_i_k_e_y_payload_s_p_param.rs
```diff
@@ -12,7 +12,7 @@ use ffi;
 pub struct MIKEYPayloadSPParam(ffi::GstMIKEYPayloadSPParam);
 
 impl MIKEYPayloadSPParam {
-    pub fn new(type_: u8, len: u8, val: u8) -> MIKEYPayloadSPParam {
+    pub fn new(type_: u8, len: u8, mut val: u8) -> MIKEYPayloadSPParam {
```
sdroege: This is not a simple `u8` but a `&[u8]`. I don't think we need a constructor at all for this as `gst_mikey_payload_sp_add_param()` internally creates it anyway.

gstreamer-sdp/src/m_i_k_e_y_payload_s_p_param.rs
```diff
@@ -29,6 +29,8 @@ impl MIKEYPayloadSPParam {
     }
 
     pub fn val(&self) -> u8 {
-        *self.0.val
+        unsafe {
+            *self.0.val
```
sdroege: This should use `slice::from_raw_parts` and return a `&[u8]`

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -28,8 +28,14 @@ glib_wrapper! {
     pub struct SDPMedia(Boxed<ffi::GstSDPMedia>);
 
     match fn {
-        copy => |ptr| ffi::gst_sdp_media_copy(mut_override(ptr)),
-        free => |ptr| ffi::gst_sdp_media_free(ptr),
+        copy => |ptr| {
+            let mut copy = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMedia>()) as *mut ffi::GstSDPMedia;
+            ffi::gst_sdp_media_copy(ptr as *const ffi::GstSDPMedia, &mut copy);
+            copy
```
sdroege: This is actually wrong, sorry.
```rust
            let mut copy = ptr::null_mut();
            ffi::gst_sdp_media_copy(ptr as *const ffi::GstSDPMedia, &mut copy);
            copy
```

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -465,7 +471,7 @@ impl SDPMedia {
         }
     }
 
-    pub fn set_media_from_caps(caps: &gst::Caps, media: SDPMedia) -> Result<(), ()> {
+    pub fn set_media_from_caps(caps: &gst::Caps, mut media: SDPMedia) -> Result<(), ()> {
```
sdroege: This should take the media by reference. We fill the existing media with information from the caps.

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -667,7 +668,7 @@ impl SDPMessage {
         assert_initialized_main_thread!();
         unsafe {
             let size = data.len() as u32;
-            let mut msg = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMessage>()) as *mut ffi::GstSDPMessage;
+            let msg = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMessage>()) as *mut ffi::GstSDPMessage;
```
sdroege: This should use `g_malloc0`, also in the other places

sdroege: I guess most of the `mem::uninitialized` here should be `mem::zeroed` instead

gstreamer-sdp/src/m_i_k_e_y_payload_s_p_param.rs
notbdu: Ah, dammit I still have to get used to the idea of a u8 pointer actually being an array/slice instead of a single u8... t.t

gstreamer-sdp/src/s_d_p_media.rs
notbdu: Curious, when is it idiomatic to use the following different methods for passing memory to a `ffi` "constructor" method. After observing the usage of each I haven't made any sensible conclusions so far... XD

  - `mem::uninitialized()`
  - `mem::zeroed()` (It looks like this method is always preferred over `mem::uninitialized`? It's possible for uninitialized memory to contain garbage?)
  - `ptr::null_mut()`
  - `g_malloc0` (After reading docs [here](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-malloc) I can see why this is preferred)

sdroege: `uninitialized`/`g_malloc` are fine if the memory is going to be fully initialized. `zeroed`/`g_malloc0` otherwise. The `mem::` functions for stack allocations, `g_malloc` for heap allocation.

sdroege: Oh and I don't really trust the SDP library to properly initialize everything. It seems to require you to zero-initialize things. As said, uncommon and annoying API :)

notbdu:
Gotcha, that makes sense to me. Thanks!!

gstreamer-sdp/src/s_d_p_media.rs
notbdu: Hmmmm, what is the rationale for using `ptr::null_mut()` here versus `g_malloc`. Both methods are allocating memory on the heap?

sdroege: First is on the stack, second on the heap. The first here because the function will allocate for us and we only need to provide a pointer where it can write to

gstreamer-sdp/src/m_i_k_e_y_payload.rs
sdroege: Still missing, this can't really compile as-is, can it?

gstreamer-sdp/src/s_d_p_connection.rs
```diff
@@ -10,17 +10,18 @@ use std::mem;
...
 
 use ffi;
+use glib::translate::*;
 
 use auto::SDPResult;
 
-
-pub struct SDPConnection(ffi::GstSDPConnection);
+#[repr(C)]
+pub struct SDPConnection(pub ffi::GstSDPConnection);
```
sdroege: IMHO these wrapper structs should not make their content `pub`. If anything, make them `pub(crate)` as needed

gstreamer-sdp/src/s_d_p_media.rs
```diff
@@ -25,55 +28,66 @@ glib_wrapper! {
...
 }
 
 impl SDPMedia {
-    pub fn add_attribute<'a, P: Into<Option<&'a str>>>(&mut self, key: &str, value: P) -> Result<_, ()> {
+    pub fn new() -> (SDPResult, SDPMedia) {
+        assert_initialized_main_thread!();
+        unsafe {
+            let mut media = ptr::null_mut();
+            let ret = from_glib(ffi::gst_sdp_media_new(&mut media));
+            (ret, from_glib_full(media))
```
sdroege: This should return a `Result`, not a tuple

gstreamer-sdp/src/s_d_p_message.rs
```diff
@@ -588,27 +664,27 @@ impl SDPMessage {
...
             let size = data.len() as u32;
-            let mut msg = glib_ffi::g_malloc(mem::size_of::<ffi::GstSDPMessage>());
-            let result = from_glib(ffi::gst_sdp_message_parse_buffer(data.to_glib_none().0, size, &mut msg));
+            let msg = glib_ffi::g_malloc0(mem::size_of::<ffi::GstSDPMessage>()) as *mut ffi::GstSDPMessage;
+            let result = from_glib(ffi::gst_sdp_message_parse_buffer(data.to_glib_none().0, size, msg));
             match result {
-                SDPResult::ok => Some(from_glib_full(msg)),
-                _ => Err(result),
+                SDPResult::Ok => Ok(from_glib_full(msg)),
+                _ => Err(()),
```
sdroege: On error you need to `glib_ffi::g_free` the pointer, everywhere where `g_malloc` is used

sdroege:
If you rebase on the `webrtc-sdp` branch now (and make sure that only my `WebRTC/SDP WIP` commit stays, none of the others from me), it should even build on travis.

gstreamer-sdp/src/s_d_p_connection.rs
notbdu: Oh nice! Didn't know about `pub(crate)`.

gstreamer-sdp/src/s_d_p_message.rs
notbdu: That is true...

sdroege:
This still does not build. Did you address all the review comments though?

sdroege:
You did, thanks :) Looks mostly good to me and I'll fix the remaining build problems and do some minor cleanup. The build problems are my fault actually

sdroege:
And all merged, thanks a lot :) I did some minor cleanup, and will do a second go over the API again tomorrow.

PR #102, opened 2018-05-05
Print pointer value when formatting debug strings

thiblahute:


gstreamer-video/src/video_info.rs
```diff
@@ -190,6 +190,14 @@ impl ::VideoMultiviewFramePacking {
 
 pub struct VideoInfo(pub(crate) ffi::GstVideoInfo);
 
+impl fmt::Debug for VideoInfo {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?} - ({:?}x{:?} - {:?}bytes)", self.format(),
+            self.width(), self.height(), self.size())
```
sdroege: This should use the `debug_struct` API like the others

gstreamer/src/segment.rs
```diff
@@ -496,13 +496,16 @@ impl<T: FormattedValue> AsRef<Segment> for FormattedSegment<T> {
 impl<T: FormattedValue> fmt::Debug for FormattedSegment<T> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let segment = self.as_ref();
+        let ptr = self as *const _;
         match segment.get_format() {
             Format::Undefined => f.debug_struct("Segment")
+                .field("ptr", &ptr)
```
sdroege: Printing the pointer here is not very useful, as the segment is stack allocated and moved around everywhere. So this value is going to change

gstreamer/src/toc.rs
```diff
@@ -96,6 +96,7 @@ impl ToOwned for TocRef {
 impl fmt::Debug for TocRef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Toc")
+            .field("ptr", unsafe { &self.as_ptr() } )
```
sdroege: There's also TocEntry here

gstreamer/src/event.rs
```diff
@@ -440,6 +440,7 @@ impl glib::types::StaticType for EventRef {
 impl fmt::Debug for EventRef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Event")
+            .field("ptr", unsafe { &self.as_ptr() } )
```
sdroege: You're missing caps also

thiblahute: It uses `self.to_string()` which should do the right thing.

gstreamer/src/toc.rs
thiblahute: Done.

gstreamer/src/segment.rs
thiblahute: Removed.

gstreamer-video/src/video_info.rs
thiblahute: Done.

gstreamer/src/event.rs
sdroege: I mean the caps `Debug` impl does not print the pointer :) We should add this consistently for all miniobjects

gstreamer-video/src/video_info.rs
```diff
@@ -190,6 +190,17 @@ impl ::VideoMultiviewFramePacking {
 
 pub struct VideoInfo(pub(crate) ffi::GstVideoInfo);
 
+impl fmt::Debug for VideoInfo {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        f.debug_struct("VideoInfo")
+            .field("format", &self.format())
+            .field("width", &self.width())
+            .field("height", &self.height())
+            .field("size", &self.size()).finish()
```
sdroege: finish in a new line

gstreamer/src/event.rs
thiblahute: This is how it is implemented:

``` rs
impl fmt::Debug for CapsRef {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        f.debug_tuple("Caps").field(&self.to_string()).finish()
    }
}
```

You want to make the imple use debug_struct you mean?

gstreamer-video/src/video_info.rs
thiblahute: Done.

gstreamer/src/event.rs
sdroege: Oh no, I was confused. Sorry :) The way how it was before is nicer IMHO. Will have to think about all this for a bit

sdroege:
I believe for things that are actually travelling in the pipeline (buffers, buffer lists, queries, events, messages) printing the pointer makes sense as that allows you to find the same thing easily in debug logs by pointer value. For other mini objects and `GstStructure` this does not have much value.

What do you think @thiblahute? What was your original use-case here?

thiblahute:
>What do you think @thiblahute? What was your original use-case here?

It was for a buffer iirc, I should just remove for ToC then?

sdroege:
Caps, sample, TOC and toc entry, yes

thiblahute:
Well, for sample I think it is useful to be able to follow it in the base class (happened quite a few time in my recent debugging). For caps, we don't use that anyway.

thiblahute:
> Caps, sample, TOC and toc entry, yes

Ok, did that, let's stop bikeshedding! :-)

PR #106, opened 2018-05-24
Tags & TOC: serialization / deserialization

fengalin:
This is an implementation of the [`serde`](https://crates.io/crates/serde) `Traits` for GStreamer's metadata.


fengalin:
The first commit is just a prototype for serialization in order to get your opinion.

gstreamer/src/tags.rs
```diff
@@ -554,6 +563,72 @@ impl StaticType for TagListRef {
     }
 }
 
+#[cfg(feature = "serde-support")]
+struct TagValue<'a>(&'a str, Value);
+#[cfg(feature = "serde-support")]
+impl<'a> Serialize for TagValue<'a> {
```
fengalin: I'm pretty sure it should be possible to use `Tag` directly here. Will try now.

fengalin:
I think I have the overall architecture. I need to figure out the specific type representations now, but it shouldn't be to much of a hassle.

sdroege:
For taglists, you can also use the serialization mechanism from GStreamer btw. They can be deserialized from the `to_string` output.

gstreamer/Cargo.toml
```diff
@@ -37,6 +42,7 @@ embed-lgpl-docs = ["rustdoc-stripper"]
 purge-lgpl-docs = ["rustdoc-stripper"]
 dox = ["gstreamer-sys/dox", "glib/dox", "futures"]
 futures = ["futures-core"]
+serde-support = ["serde"]
```
sdroege: You don't need to add a feature but could directly use `serde` AFAIU

gstreamer/src/tags.rs
```diff
@@ -17,6 +17,22 @@ use glib::translate::{from_glib, from_glib_full, ToGlib, ToGlibPtr, ToGlibPtrMut
...
+use serde::ser::{Serialize, Serializer, SerializeSeq, SerializeTuple};
+#[cfg(feature = "serde-support")]
+use serde::de;
+#[cfg(feature = "serde-support")]
+use serde::de::{Deserialize, Deserializer, SeqAccess, Visitor};
+#[cfg(feature = "serde-support")]
+use glib::translate::from_glib_none;
+#[cfg(feature = "serde-support")]
+use gobject_ffi;
+
```
sdroege: It might be nicer to move this into a submodule inside this file, then you only need to use the `#cfg` once for the module

gstreamer/src/tags.rs
```diff
@@ -554,6 +570,179 @@ impl StaticType for TagListRef {
...
+#[cfg(feature = "serde-support")]
+impl<'a> Serialize for TagSer<'a> {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut tup = serializer.serialize_tuple(2)?;
+        tup.serialize_element(self.0)?;
+        match self.1.type_() {
+            // TODO: use a macro serialize_tag_value!
+            glib::Type::String => tup.serialize_element(&self.1.get::<String>().unwrap())?,
+            glib::Type::U32 => tup.serialize_element(&self.1.get::<u32>().unwrap())?,
```
sdroege: There's also `gst::ValueExt::serialize` and the inverse which could be useful here

gstreamer/src/tags.rs
```diff
@@ -554,6 +570,179 @@ impl StaticType for TagListRef {
...
+            }
+        }
+        tup.end()
+    }
+}
+
+#[cfg(feature = "serde-support")]
+impl Serialize for TagList {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let tag_count = unsafe { ffi::gst_tag_list_n_tags(self.as_ptr()) };
```
sdroege: @ystreet wanted an API to be able to iterate over all tags for reading and having them as `Value`s then. This would simplify this code here a lot

gstreamer/src/tags.rs
```diff
@@ -554,6 +570,179 @@ impl StaticType for TagListRef {
...
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("a sequence of Tags")
+    }
+
+    fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
+        let tag_list = TagList::new();
+        while let Some(tag_de) = seq.next_element::<TagDe>()? {
+            unsafe {
+                ffi::gst_tag_list_add_value(
```
sdroege: There should probably be some checking if the value is of the correct type for the tag here

sdroege:
Generally looks nice but I believe I'd want to have some generic tag iteration/getter API, and maybe even a setter around `Values`. The setter would have to check the tag value types though.

This should simplify your code a lot and remove all the unsafety from it

fengalin:
Thanks for the review! I'll get through this later.

I noticed the `to_string` methods, but since `serde` supports different formats, I though it was worth the implementation.

gstreamer/Cargo.toml
fengalin: That would be better indeed.

gstreamer/src/tags.rs
fengalin: I'll do that. Thanks.

gstreamer/src/tags.rs
fengalin: Yes, that would help.

sdroege:
> but since serde supports different formats, I though it was worth the implementation.

I agree, yes

gstreamer/src/tags.rs
ystreet: My current WIP is at https://github.com/sdroege/gstreamer-rs/pull/107 for this :)

fengalin:
I implemented `serde` for the "complex" type `DateTime`. A few notes regarding my choices:
- I used a different feature name for `serde` again because I needed to pull `serde_derive` also and couldn't find a way to make `serde` depend on `serde_derive`. If there is a solution, I'd be glad to implement it.
- Serialization and deserialization of `DateTime` don't rely on the same intermediate type (resp. `DateTimeSe` and `DateTimeDe`). I separated them because:
    * During deserialization, I need all the fields to be read before calling `DateTime::new`. The easiest way to do that was to implement a `struct` and annotate it with `Deserialize` (hence the need for `serde_derive`). 
    * During serialization, I wanted to avoid allocating an intermediate `struct` with all the fields.
- AFAICT, the tag's type is controlled during deserialization thanks to the [`TagVisitor`](https://github.com/fengalin/gstreamer-rs/blob/545aaa39cff5fc10baea92f97a0e9246283c8f1c/gstreamer/src/tags.rs#L773).

Please tell me if you think this is right. I only need to add support for `Sample` then. Except for `serde`'s buffer handling specificities, the rest of the code should be similar to what I did for `DateTime`.

Note: CI's `curl` command failed for `beta` channel. Could you restart it when you have time please?

gstreamer/src/tags.rs
```diff
@@ -594,6 +595,8 @@ mod serde {
                 type_ => {
                     if DateTime::static_type() == type_ {
                         ser_tag!(serializer, self, DateTime)
+                    } else if Sample::static_type() == type_ {
+                        ser_tag!(serializer, self, DateTime)
```
sdroege: This should be `Sample` :)

fengalin: Oups! :) I didn't implement it yet anyway. Will be available soon.

fengalin:
Cleaned up the commit set...
Implementing serde for other types in order to be able to serde Sample. There are more of them than I thought...

fengalin:
Buffer's byte stream representation is not efficient with `ron` and `json`. We might need a means to prevent the buffer from being exported or to select `base64` encoding.

sdroege:
> Buffer's byte stream representation is not efficient with ron and json. We might need a means to prevent the buffer from being exported or to select base64 encoding.

Why? If you let ron/json directly serialize binary data, what does it do? Something else than base64 encoding as a string?

fengalin:
See for yourself [`test_serialize` for `Buffer`](https://github.com/fengalin/gstreamer-rs/blob/3889d86febe9cba88e40e2e1497fb99bd6e16e37/gstreamer/src/buffer.rs#L737). I've included `ron`, `json` and `pickle` so that we can see what they look like. Please note that I used `serde_bytes` which takes care of optimizing byte array representations when possible.

My position now is that the user should be able to select the binary representation for text formats. This would be similar to what `ron` provides with `PrettyConfig` (see [this](https://github.com/fengalin/gstreamer-rs/blob/3889d86febe9cba88e40e2e1497fb99bd6e16e37/gstreamer/src/buffer.rs#L758)). In other words, this would be an evolution on `ron` & `serde_json` crates.

sdroege:
The `pickle` variant seems fine, the others are a bit verbose indeed. I would consider that to be a `serde` problem (or rather of the `rson`/`json` crates). So fine for me to keep it like this as long as you report this problem there also :)

fengalin:
I've just sent https://github.com/ron-rs/ron/pull/109 with the changes. I will do the same for `serde_json` later. Good news: there's no need for a new configuration entry. That's what `serde_bytes` is designed for.

fengalin:
I have `serde` implemented and tested on all tag types now. Please note the following comments, limitations and issues though:

- `ron` team seems happy with the [enhancement I proposed](https://github.com/ron-rs/ron/pull/109) and they said they will release a new version. In the meantime, `Cargo.toml` points to my branch.
- I need some time to figure out how to optimize buffer serialization with `serde_json`. We can't just use base64 with this format as this is not part of the "standard" AFAICT. Nonetheless, It doesn't seem like the `serde_bytes` stuff is used. I will perform some more tests and report upstream when I get my ideas clear. It might take some time as I resume paid job on Monday :( and I am busy this week-end.
- I didn't implement `serde` traits for `Segment` yet. The reason is that I don't think it was necessary for `Tags` and it required implementing `serde` on all the variants of `GenericFormattedValue`. This shouldn't be a problem but it will take time... Serializing a `Sample` [generates a `struct`](https://github.com/fengalin/gstreamer-rs/blob/da72650ca8928771e07607c956465cc5f5a47ba3/gstreamer/src/sample.rs#L140) which includes the field `segment`, but it will always be `None`. I decided to include it so that the resulting representation remains compliant with a future support for `Segment`.
- I had hard times dealing with `Structure`. I would like to investigate further, but here again it might take some time. Here are my observations so far:
  * They are memory issues visible executing [`sample::tests::test_serialize`](https://github.com/fengalin/gstreamer-rs/blob/da72650ca8928771e07607c956465cc5f5a47ba3/gstreamer/src/sample.rs#L221). (Message: `GStreamer-CRITICAL **: 21:58:17.219: gst_structure_free: assertion 'GST_STRUCTURE_REFCOUNT (structure) == NULL' failed`) and they are related to using the `info` with the `Sample`. This is the least problematic situation I could reach. The worse being seen moving the [`info` instantiation](https://github.com/fengalin/gstreamer-rs/blob/da72650ca8928771e07607c956465cc5f5a47ba3/gstreamer/src/sample.rs#L247) before [the `sample` scope](https://github.com/fengalin/gstreamer-rs/blob/da72650ca8928771e07607c956465cc5f5a47ba3/gstreamer/src/sample.rs#L232), in which case, it produces a `signal: 11, SIGSEGV: invalid memory reference`. I couldn't find any obvious reason for this yet.
  * Dealing with a `Structure` as part of a `Sample` instantiation can be cumbersome. In my case, I had to get an `Option<&StructureRef>` from an `Option<Structure>`. I might have missed something, but the [best code I figured out](https://github.com/fengalin/gstreamer-rs/blob/da72650ca8928771e07607c956465cc5f5a47ba3/gstreamer/src/sample.rs#L175) for this doesn't seem right to me.
- I still need to use a more optimal way to [`match` the `Value` types](https://github.com/fengalin/gstreamer-rs/blob/da72650ca8928771e07607c956465cc5f5a47ba3/gstreamer/src/tags.rs#L590) than using `glib::Type`.
- I really would like to implement `Toc` support as this was my initial goal. Hopefully I'll be able to do this on Sunday.
- For some reasons, CI doesn't execute my `ser_de` feature gated tests even with `cargo test --all --all-features`... Running `cargo test --all-features` locally works OK.

sdroege:
> ron team seems happy with the enhancement I proposed and they said they will release a new version. In the meantime, Cargo.toml points to my branch.

Great, thanks for caring for this :) We should wait until they merged/released it before merging this here though, otherwise it might potentially block me from making releases

> I need some time to figure out how to optimize buffer serialization with serde_json. We can't just use base64 with this format as this is not part of the "standard" AFAICT. Nonetheless, It doesn't seem like the serde_bytes stuff is used. I will perform some more tests and report upstream when I get my ideas clear. It might take some time as I resume paid job on Monday :( and I am busy this week-end.


Take your time :) But this is only relevant for the tests here right now, and an inherent limitation of JSON / `serde-json` currently?

>  didn't implement serde traits for Segment yet. The reason is that I don't think it was necessary for Tags and it required implementing serde on all the variants of GenericFormattedValue. This shouldn't be a problem but it will take time... Serializing a Sample generates a struct which includes the field segment, but it will always be None. I decided to include it so that the resulting representation remains compliant with a future support for Segment.

That seems fine. You could alternatively also store the segment how it is stored in C (check the definition in `gstreamer-sys`), that should be much easier for serialization and is equivalent. Same goes for `GenericFormattedValue`, just store that as a pair of format and value.

> I had hard times dealing with Structure. I would like to investigate further, but here again it might take some time.

Generally it should be stored like a dictionary plus a name in the front. Keys are strings, values are "values". Each value can be *anything* you can store in a `GValue`, but in practice it's going to be integers, strings, etc.. I would suggest that you implement direct support for the common types (I can give you a list), and otherwise fall back to `gst::ValueExt::serialize`

> They are memory issues visible executing sample::tests::test_serialize. (Message: GStreamer-CRITICAL **: 21:58:17.219: gst_structure_free: assertion 'GST_STRUCTURE_REFCOUNT (structure) == NULL' failed) and they are related to using the info with the Sample. This is the least problematic situation I could reach. The worse being seen moving the info instantiation before the sample scope, in which case, it produces a signal: 11, SIGSEGV: invalid memory reference. I couldn't find any obvious reason for this yet.

I'll take a look into that, seems bad and possibly a memory unsafety in the bindings somewhere.

> Dealing with a Structure as part of a Sample instantiation can be cumbersome. In my case, I had to get an Option<&StructureRef> from an Option<Structure>. I might have missed something, but the best code I figured out for this doesn't seem right to me.

You should be able to always work on `&StuctureRef` for serialization and `Structure` for deserialization. It's basically like `&Path` and `PathBuf`, or `&str` and `String`.

> I still need to use a more optimal way to match the Value types than using glib::Type.

What were you thinking of exactly? I would match the common types somehow like you do, and for everything else fall back to `gst::ValueExt::serialize`.

> For some reasons, CI doesn't execute my ser_de feature gated tests even with cargo test --all --all-features... Running cargo test --all-features locally works OK.

We should change the `.travis.yml` to not use `--all-features`. It easily breaks. Unfortunately setting features when building a workspace crate does not work well, so we would have to build each crate individually and pass `--feature` for the features we want to test (none, `v1_14` at least).

fengalin:
> But this is only relevant for the tests here right now, and an inherent limitation of JSON / serde-json currently?

Yes, at least that's what I think. I'll make sure when I get time.

> [about Segment] You could alternatively also store the segment how it is stored in C

Yes indeed, it seems a lot easier viewed as a [C struct](https://developer.gnome.org/gstreamer/unstable/gstreamer-GstSegment.html#GstSegment)

> [About Structure] Each value can be anything you can store in a GValue, but in practice it's going to be integers, strings, etc.. I would suggest that you implement direct support for the common types (I can give you a list), and otherwise fall back to gst::ValueExt::serialize

I checked [glib::types::Type](http://gtk-rs.org/docs/glib/types/enum.Type.html) before. The first part are trivial primitive types. I guess it gets tricky from `BaseInterface`, but I also guess this is the kind of stuff that could fallback to using `gst::ValueExt::serialize`. By the way, what would you think of implementing the `glib::value::Value` serialization (until `Variant` included) in the `glib` crate? I don't want to extend the scope of this PR too much, but that could be a nice base for implementing `serde` in other crates.

> You should be able to always work on &StuctureRef for serialization and Structure for deserialization. It's basically like &Path and PathBuf, or &str and String

That's what I did. However, constructing a `Sample` [requires an `Option<&StructureRef>`](https://github.com/fengalin/gstreamer-rs/blob/1192a027b40c5ad46cd998cfa7fd11bc527bd1a9/gstreamer/src/sample.rs#L34). When I reach this in deserialization, I handle an `Option<Structure>` so I have to match on the Option's variants in [this inelegant way](https://github.com/fengalin/gstreamer-rs/blob/da72650ca8928771e07607c956465cc5f5a47ba3/gstreamer/src/sample.rs#L175). It would be convenient if we add another `new_...` function which would take an `Option<Structure>`. Maybe some kind of magic could be possible with an `Option<AsRef<>>` in the `new` signature somewhere. I'll try and experiment with that.

> [about matching the `Value` types] What were you thinking of exactly? I would match the common types somehow like you do, and for everything else fall back to gst::ValueExt::serialize.

I was referring to [this previous comment](https://github.com/sdroege/gstreamer-rs/pull/107#issuecomment-393127118). It was merely to note that I didn't forget about it.

sdroege:
> By the way, what would you think of implementing the glib::value::Value serialization (until Variant included) in the glib crate? I don't want to extend the scope of this PR too much, but that could be a nice base to implement serde in other crates.

Unfortunately there is no serialization API available in GLib, that's why we added that as an extension in GStreamer.

Adding custom serialization to `glib-rs` for `GValues` might be a contentious topic. You could create an issue and check what people think about it.

Also that it will never cover all types seems suboptimal, and e.g. you won't be able to cover types that GStreamer **can** serialize but GLib does not know about (`GstFraction`, the ranges, etc).

> Maybe some kind of magic could be possible with an Option<AsRef<>> in the new signature somewhere. I'll try and experiment with that.

Using `T: Option<AsRef<StructureRef>>` probably works, to be tried :)

fengalin:
> Adding custom serialization to glib-rs for GValues might be a contentious topic. You could create an issue and check what people think about it.

I investigated a little further. The issue I find so far has to do with types such as `Type::Pointer`, `Type::BaseInterface`, etc. AFAIU, `glib` provides introspection mechanisms, so I guess those could be used to dynamically "expand" the actual types during ser/de. I think I can support my suggestion in `gtk-rs/glib` if I have a proposal for this. Do you think using introspection makes sense, or would it be suboptimal?

> Also that it will never cover all types seems suboptimal, and e.g. you won't be able to cover types that GStreamer can serialize but GLib does not know about (GstFraction, the ranges, etc).

My idea is that `gstreamer-rs` could check its own types first and then pass `serde` handling to `glib::Value` for others.

fengalin:
> It seems quite complicated and not sure if it's worth the effort.

That's what I was afraid of. I will stay on GStreamer side for this then. Thank you for the insights.

sdroege:
> That's what I was afraid of. I will stay on GStreamer side for this then. Thank you for the insights.

Sorry for not having a more useful answer :)

fengalin:
Don't be, that was really useful and I suspected something potentially too complicated.

fengalin:
It would be convenient if we could instruct `Gir` to generate `serde` derives, yielding something like this:
``` rust
#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
#[cfg(feature = "ser_de")]
#[derive(Serialize, Deserialize)]
pub enum Format {
    Undefined,
    [...]
}
```

I was hoping for something like this:
``` toml
[[object]]
name = "Gst.Format"
status = "generate"
cfg_condition = "feature = \"ser_de\""
    derive = "Serialize, Deserialize"
```

Is there an option to achieve this or should I rely on a manual implementation instead? Do you think that's something we should add to `Gir`?

sdroege:
That seems like a useful idea, at least for serde. But it would also only really work for enums/flags types, nothing else.

fengalin:
For the record: https://github.com/gtk-rs/gir/pull/611

fengalin:
Last updates:

- Reworked the commit set.
- Implemented serde for types in `values.rs` (except the collections of `glib::SendValue`s).
- Simplified `GenericFormattedValue` & `Buffer` serde impl thanks to gtk-rs/gir#611.

TBD:

- `Segment` serde impl.
- Optimize `glib::Value::type` `match` arms.
- Update `ron` version when it is released (ron-rs/ron#109 is merged).

sdroege:
What do you mean with collections of `glib::SendValue`?

fengalin:
I was referring to structs such as https://github.com/sdroege/gstreamer-rs/blob/bd1fb746de610a19835790c2ead81eac994a336b/gstreamer/src/value.rs#L525

I'm pretty sure I can support these too, but I need to dig a little more.

sdroege:
Ah for those, you would basically serialize them like a vector (and always deserialize to an owned variant) and remember the concrete type (array or list)

fengalin:
I think I have it all together now, except for the `static_type`s comparison: https://github.com/fengalin/gstreamer-rs/blob/5d34ce1f325a904b43a79eb00d787e997a86872f/gstreamer/src/value.rs#L925

I couldn't find any ffi enum for these, so the only way I can think of would be to use a `lazy_static`. Do you think about any other way?

I have to say I'm not too happy with the `Array`s when they are part of a `Structure`: https://github.com/fengalin/gstreamer-rs/blob/56ad173fb02062d1037080622d22025e1078e8ff/gstreamer/src/structure.rs#L909

There would be an easy way to make it look like this:
```
("array", "Array", ([
    ("i32", 1),
    ("i32", 2),
])),
```
but I think it'd be a bit weird to have the array as a single item in a tuple `([ ... ])`. What do you think?

Another thing I'm not too happy with is the way I have to deserialize the entire `Structure` field list before adding them one by one to the target `Structure`: https://github.com/fengalin/gstreamer-rs/blob/56ad173fb02062d1037080622d22025e1078e8ff/gstreamer/src/structure.rs#L770

I couldn't find a way to avoid this.

Other than that, all the TBD items I mentioned previously are handled now (including depending on the `ron` release) and the unit tests are finally executed on CI \o/

fengalin:
This time I think the implementation is complete now. Here are a few points I'd like to discuss with you.

#### `StaticType` comparisons

I used `lazy_static`s to avoid calling `static_type()` for types for which I couldn't find any matching `ffi` `enum`. [See this](https://github.com/fengalin/gstreamer-rs/blob/8c9edc83279d8f6a537a65b5f2d3aa8508591330/gstreamer/src/value.rs#L939)

Tell me if you think of a better way to handle this.

#### `Segment` within a `Sample`

When initializing a `Sample` with a `None` segment, [`GStreamer` currently allocates the actual `Segment`](https://github.com/GStreamer/gstreamer/blob/cd3e55c38e2ba1d7cb1e0565da302d7897b3b70b/gst/gstsample.c#L126). This should be changed as part of version 2.0. In the meantime, such a `Sample` is serialized with the initialized `Segment`, which is "a bit" verbose, especially considering that `Sample` is used to store images in `Tag`s. As a work around, I propose to detect when the `Sample::get_segment()` returns a `Segment` with the defaults and serialize it as `None` is that case. This is the current serialization of a `Sample` initialized with a `None` `Segment`:

```
(
    buffer: Some((
        pts: Some(1),
        dts: None,
        duration: Some(4),
        offset: 0,
        offset_end: 4,
        flags: (
            bits: 0,
        ),
        buffer: "AQIDBA==",
    )),
    buffer_list: None,
    caps: Some([
        ("sample/caps", [
            ("my_int", "i32", 12),
            ("my_bool", "bool", true),
        ]),
    ]),
    segment: Some((
        flags: (
            bits: 0,
        ),
        rate: 1,
        applied_rate: 1,
        format: Time,
        base: Time(Some(0)),
        offset: Time(Some(0)),
        start: Time(Some(0)),
        stop: Time(None),
        time: Time(Some(0)),
        position: Time(Some(0)),
        duration: Time(None),
    )),
    info: Some(("sample.info", [
        ("f3", "i32", 123),
    ])),
)
```

sdroege:
> I used lazy_statics to avoid calling static_type() for types for which I couldn't find any matching ffi enum. See this

Why would you use `lazy_static!` and just call the corresponding functions?



> As a workaround, I propose to detect when the Sample::get_segment() returns a Segment with the defaults and serialize it as None is that case. This is the current serialization of a Sample initialized with a None Segment

An actual `None` segment would be with `gst::Format::Undefined` and that we could serialize easily. It's quite wrong that `GstSample` is initializing a `Time` segment if nothing is given :) It would be the exactly same segment that you would initialize the sample with in most other pipelines too (before a seek), so currently there`s no way to distinguish the two cases.

I'd keep it as verbose as it is but I would serialize the segment just like the underlying C struct. That is, you serialize the format once and for all the fields you directly write the underlying 64 bit integers instead of the formatted ones (getting rid of the duplicated `Time` and `Some`/`None` is probably simpliyfing it quite a bit already).

fengalin:
> Why would you use lazy_static! and not just call the corresponding functions?

Because I remembered you said something about the type comparisons being suboptimal ;) so I was looking for a way to optimize by avoiding a new function call for repeated checks for the same type. But looking back at [your comment](https://github.com/sdroege/gstreamer-rs/pull/107#issuecomment-393127118), it seems you were concerned by using the `glib::Type` instead of the matching `usize`, which I don't get as I though, being an `enum`, [`glib::Type`](http://gtk-rs.org/docs/glib/types/enum.Type.html) was already pretty optimized.

> I'd keep it as verbose as it is but I would serialize the segment just like the underlying C struct. That is, you serialize the format once and for all the fields you directly write the underlying 64 bit integers instead of the formatted ones (getting rid of the duplicated Time and Some/None is probably simpliyfing it quite a bit already).

Don't you think representing `None` variants has value here? For a default `Segment`, `stop` and `duration` are set to `None` which is a different information than `0` IMHO.

I agree that we could avoid repeating `Time()` for each value providing that `format` already gives this information if all of {`base`, `offset`, `start`, `stop`, `time`, `position` and `duration`} are guaranteed to match this very type. I'm relying on the generic auto generated ser/de mechanisms for `GenericFormattedValue` here which is quite convenient. I'll see what I can do.

sdroege:
> it seems you were concerned by using the glib::Type instead of the matching usize, which I don't get as I though, being an enum, glib::Type was already pretty optimized.

It's two comparisons except of one if it was just a single integer for all non-basic types. Not really much of a problem

> are set to None which is a different information than 0 IMHO.

None is -1 and that has no additional meaning. It seems shorter for the serialization and provides the same amount of information

What I meant is serializing basically the C struct, not the more complicated type we do in Rust

fengalin:
> It's two comparisons except of one if it was just a single integer for all non-basic types. Not really much of a problem

Got it!

> None is -1 and that has no additional meaning. It seems shorter for the serialization and provides the same amount of information

All right! Only problem left is that [`GstSegment` fields](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstSegment.html#GstSegment-struct) are all `guint64`, not `gint64`, so the C `-1` is serialized as `18446744073709551615`. I see that only `stop` and `duration` seem concerned with a potential `-1`, so I suggest keeping the `Option` for these and use the `guint64` value directly for the others.

fengalin:
I think it's ready for review now.

fengalin:
When I merge with PR #119, I'll gather the tags by name, something like this:
```
[
    ("title", [
        "a title",
        "another title",
    ]),
...
]
```

sdroege:
> so the C -1 is serialized as 18446744073709551615. I see that only stop and duration seem concerned with a potential -1, so I suggest keeping the Option for these and use the guint64 value directly for the others.

I would prefer to use the actual underlying integers for all of them.

> When I merge with PR #119, I'll gather the tags by name, something like this:

That looks good :) I'd prefer to get that PR in first in any case, it would simplify some code here

gstreamer/src/buffer.rs
```diff
@@ -579,6 +579,79 @@ lazy_static! {
...
+    use serde_bytes::{Bytes, ByteBuf};
+
+    use Buffer;
+    use BufferFlags;
+    use BufferRef;
+    use ClockTime;
+
+    impl<'a> Serialize for BufferRef {
+        fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+            let mut datetime = serializer.serialize_struct("Buffer", 6)?;
```
sdroege: datetime?

gstreamer/src/buffer.rs
```diff
@@ -579,6 +579,79 @@ lazy_static! {
...
+            datetime.serialize_field("pts", &self.get_pts())?;
+            datetime.serialize_field("dts", &self.get_dts())?;
+            datetime.serialize_field("duration", &self.get_duration())?;
+            datetime.serialize_field("offset", &self.get_offset())?;
+            datetime.serialize_field("offset_end", &self.get_offset_end())?;
+            datetime.serialize_field("flags", &self.get_flags())?;
+            {
+                let data = self.map_readable().unwrap();
```
sdroege: Instead of panicking, this should fail cleanly I guess

gstreamer/src/buffer.rs
```diff
@@ -579,6 +579,79 @@ lazy_static! {
...
+        buffer: ByteBuf,
+    }
+
+    impl From<BufferDe> for Buffer {
+        fn from(mut buf_de: BufferDe) -> Self {
+            // Move the `ByteBuff`'s byte array into the `gst::Buffer`
+            // Are we really avoiding copies with that?
+            let drained_buffer: &mut Vec<u8> = buf_de.buffer.as_mut();
+            let mut buffer =
+                Buffer::from_slice(drained_buffer.drain(..).collect::<Vec<u8>>()).unwrap();
```
sdroege: drain and collect will do a copy, so you're not really gaining anything here

gstreamer/src/caps.rs
```diff
@@ -397,6 +397,67 @@ impl Builder {
     }
 }
 
+#[cfg(feature = "ser_de")]
+pub(crate) mod serde {
```
sdroege: Maybe we can move those serde submodules into different files, like `src/caps_serde.rs`, and then just import them into `src/caps.rs`? One level of indentation less and it's clearer where to find the code

gstreamer/src/caps.rs
```diff
@@ -397,6 +397,67 @@ impl Builder {
...
+    use std::fmt;
+
+    use Caps;
+    use CapsRef;
+    use Structure;
+
+    impl<'a> Serialize for CapsRef {
+        fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+            let iter = self.iter();
+            let size = iter.size_hint().0;
```
sdroege: Strictly speaking, next to each structure the caps also have an optional `GstCapsFeature`. That API is not part of the bindings yet, so just add a FIXME comment about that somewhere here

gstreamer/src/toc.rs
```diff
@@ -256,6 +256,122 @@ impl fmt::Debug for TocEntryRef {
...
+    }
+
+    impl Serialize for TocEntryRef {
+        fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+            let mut toc_entry = serializer.serialize_struct("TocEntry", 6)?;
+            toc_entry.serialize_field("entry_type", &self.get_entry_type())?;
+            toc_entry.serialize_field("uid", &self.get_uid())?;
+            toc_entry.serialize_field("start_stop", &self.get_start_stop_times())?;
+            toc_entry.serialize_field("tags", &self.get_tags())?;
+            toc_entry.serialize_field("loop_", &self.get_loop())?;
```
sdroege: Why with underscore?

sdroege:
Looks mostly good to me, thanks :)

gstreamer/src/buffer.rs
fengalin: Sorry, copy/paste :)

gstreamer/src/buffer.rs
fengalin: Sure thing!

gstreamer/src/buffer.rs
fengalin: All right, I'll see what I can do.

gstreamer/src/caps.rs
fengalin: Indeed. Some of them are getting quite large too.

gstreamer/src/caps.rs
fengalin: Would that lead to something like:
```
(
    structures: [
    ],
    features: Some([
    ]),
)
```

If so, we might want to include both fields right now and keep `features: None` until `GstCapsFeautre` is part of the binding. That would be future proof. Alternatively, we could just remove the `Option` part and just keep the array, which can be empty.

sdroege: It would be more like
```
structure: [
    (structure1, feature1),
    (structure2, feature2),
],
```

gstreamer/src/toc.rs
sdroege: Ah I thought for the actual serialization "output" (like the field names in the JSON) we could use any string we want and let serde rename that. That's not the case, or not simple in this case? With `serde-derive` there's an attribute to do that at least

fengalin: Yes, I think I saw this before. Thanks for pointing this.

fengalin:
Please see commit https://github.com/sdroege/gstreamer-rs/pull/106/commits/5c6a938fea652e111eec0e63c115272df2e0a834 as a preview of what I think is required to separate the `serde` implementations and tests from their base modules. The preview contains variations as applied to the `date_time` and `value` modules.

The "interesting" part is for `value`. AFAIU, because the `serde` implementations for types in `value` access private fields from the base module, it is necessary to introduce a submodule, which (as of Rust 2015) requires to add a `value` directory and move `value.rs` to ` value/mod.rs`. Note that in Rust 2018, it seems it will be possible to keep `value.rs` as is ([see § "no more `mod.rs`"](https://rust-lang-nursery.github.io/edition-guide/2018/transitioning/modules/path-clarity.html#no-more-modrs)).

Also, I decided to name the submodule `ser_de` instead of `value_serde` as: a) I though it was not necessary to add the `value_` prefix as it is a submodule of `value`, b) I can't name it `serde` in Rust 2015 as it's the name of a crate and I didn't find a way to indicate that I was referring to the submodule from the base module (statement `pub(crate) mod ser_de;`).

I included the variation for `date_time` which is less invasive as it doesn't require a submodule as its `serde` implementation doesn't use private fields of the base module. IMO, keeping these might seem inconsistent with other types that need a submodule.

My idea would be to go with the all submodule approach for all types with a `serde` implementation. But that leads to a lot of files moving around. What do you think? Shall I proceed with these modifications?

sdroege:
You could also mark the private fields/functions that you need to use as `pub(crate)`. Would that improve anything here?

fengalin:
Yes, that's another option and it would be the least invasive. Note that the `serde` implementations and tests wouldn't be imported  as submodules in that case though.

sdroege:
> Note that the serde implementations and tests wouldn't be imported as submodules in that case though.

What do you mean with that?

sdroege:
Oh yes, but as it only defines trait impls that is not a problem really. `lib.rs` would `use` that module but not re-export it

fengalin:
I'd like to know your opinion on the following options.

I used Rust types to represent the serialized `glib::Value`s. E.g.:

```
    ("i32", 12),
    ("bool", true),
    ("Array", [
        ("i32", 1),
        ("i32", 2),
    ]),
    ("Fraction", (1, 2)),
    ("String", "test str"),
```

I like it because it is compact and self explanatory. Another option would be to use `glib` types:

```
    ("gint", 12),
    ("gboolean", true),
    ("GstValueArray", [
        ("gint", 1),
        ("gint", 2),
    ]),
    ("GstFraction", (1, 2)),
    ("gchararray", "test str"),
```

It might seem more familiar to people who are used to GLib & GStreamer, but it is more verbose and it is ambiguous in some cases (e.g. for `gint`) IMHO.

Another question: one side case was for `glib::Type::ILong` & `glib::Type::ULong` which I mapped to `i32` and `u32` as I though that was the matching non-ambiguous types in Rust. What do you think?

I'm still working on this PR, but there are quite a few things left to be done and I don't have much bandwidth due to paid work. I'm planning on pushing a new version with the remaining modifications soon (hopefully by the end of the week-end).

sdroege:
> I'd like to know your opinion on the following options.

Sorry, I was away over the weekend

> I like it because it is compact and self explanatory.

I prefer this variant too

> Another question: one side case was for glib::Type::ILong & glib::Type::ULong which I mapped to i32 and u32 as I though that was the matching non-ambiguous types in Rust. What do you think?

`long` is 64 bit on 64 bit platforms except for Windows. The safe choice here would be to use `i64` and `u64`. Where do you need to handle `long` `GValue`s?

> There's something that I don't understand: when I read the tag name with a regular String, I get intermittent issues in ffi::gst_tag_get_type. Depending on the compilation, I can call ffi::gst_tag_get_type a couple of times before it returns NULL and a log showing:

That's probably https://github.com/sdroege/gstreamer-rs/commit/93b32882eec35546a3efd05a9e260664b2f5da78 ? :)

fengalin:
> I prefer this variant too

Great! Let's keep it like this then.

> Where do you need to handle long GValue

I don't. I just wanted to handle fondamental types. Do you think I should  remove `glib::Type::ILong` and `glib::Type::ULong`? Otherwise, I'll change the type to the 64bits variants.

> That's probably 93b3288 ? :)

Great, I'll thry this tonight.

sdroege:
> I don't. I just wanted to handle fondamental types. Do you think I should remove glib::Type::ILong and glib::Type::ULong? Otherwise, I'll change the type to the 64bits variants.

It's a bit dangerous. You could serialize a too big value on a 64 bit system and then deserialization fails on a 32 bit system.

I would add no support for them until actually needed

fengalin:
I'm not planning any new changes now. Please tell me if you find something that would require an enhancement.

sdroege:
Thanks, will do. Last time I looked it seemed pretty complete already :)

The crash with the tag types is gone?

fengalin:
\o/
Yes it is!

gstreamer/src/buffer_serde.rs
```diff
@@ -0,0 +1,223 @@
...
+#[derive(Deserialize)]
+struct BufferDe {
+    pts: ClockTime,
+    dts: ClockTime,
+    duration: ClockTime,
+    offset: u64,
+    offset_end: u64,
+    flags: BufferFlags,
+    buffer: ByteBuf,
+}
```
sdroege: One thing I was wondering is if it would make sense to use that struct also for serialization and let both directions be derived and then only map in both directions. That way the struct definition does not have to be kept in sync with the separate serialization steps above.

gstreamer/src/caps_serde.rs
```diff
@@ -0,0 +1,206 @@
...
+struct CapsVisitor;
+impl<'de> Visitor<'de> for CapsVisitor {
+    type Value = Caps;
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("a sequence of `(Structure, Option<CapsFeature>)`")
+    }
+
+    fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
+        let mut caps = Caps::new_empty();
```
sdroege: Does this work with `gst::Caps::new_empty()` vs `gst::Caps::new_any()`? Both have no structures but they're different

gstreamer/src/date_time_serde.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut datetime = serializer.serialize_struct("DateTime", 8)?;
+        datetime.serialize_field("tz_offset", &self.get_time_zone_offset())?;
+        datetime.serialize_field("y", &self.get_year())?;
+        datetime.serialize_field("m", &self.get_month())?;
+        datetime.serialize_field("d", &self.get_day())?;
+        datetime.serialize_field("h", &self.get_hour())?;
+        datetime.serialize_field("mn", &self.get_minute())?;
+        datetime.serialize_field("s", &self.get_second())?;
+        datetime.serialize_field("us", &self.get_microsecond())?;
```
sdroege: `DateTime` can have no ms, s, m, etc. Calling the getters if they don't have the field will cause an assertion, and there are the `has_XXX()` functions for checking their existence

gstreamer/src/format.rs
```diff
@@ -12,6 +12,7 @@ use ClockTime;
 use Format;
 
 #[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
+#[cfg_attr(feature = "ser_de", derive(Serialize, Deserialize))]
```
sdroege: Ah interesting, I would've just handled this like a `(gst::Format, i64)` :)

gstreamer/src/segment_serde.rs
```diff
@@ -0,0 +1,208 @@
...
+
+use serde::de::{Deserialize, Deserializer};
+use serde::ser::{Serialize, SerializeStruct, Serializer};
+
+use Format;
+use GenericFormattedValue;
+use Segment;
+use SegmentFlags;
+
+impl<'a> Serialize for Segment {
```
sdroege: Does this also allow to work with `FormattedSegment`s directly, or would one have to cast first?

gstreamer/src/value_serde.rs
```diff
@@ -0,0 +1,506 @@
...
+    pub(crate) static ref FRACTION_OTHER_TYPE_ID: usize = get_other_type_id::<Fraction>();
+    pub(crate) static ref FRACTION_RANGE_OTHER_TYPE_ID: usize =
+        get_other_type_id::<FractionRange>();
+    pub(crate) static ref INT_RANGE_I32_OTHER_TYPE_ID: usize = get_other_type_id::<IntRange<i32>>();
+    pub(crate) static ref INT_RANGE_I64_OTHER_TYPE_ID: usize = get_other_type_id::<IntRange<i64>>();
+    pub(crate) static ref LIST_OTHER_TYPE_ID: usize = get_other_type_id::<List>();
+    pub(crate) static ref SAMPLE_OTHER_TYPE_ID: usize = get_other_type_id::<Sample>();
+}
+
+impl<'a> Serialize for Fraction {
```
sdroege: This one can't be derived?

gstreamer/src/buffer_serde.rs
fengalin: I didn't do that because I wanted to avoid copies, especially for the `buffer` field. I need to check again, but I don't think I can make it borrow the buffer during ser and own it in de. But it's true that consistency would be verified at compile time with your proposal. Other types such as `DateTime` would benefit from this approach without much extra overhead though.

gstreamer/src/caps_serde.rs
fengalin: Oh, I didn't noticed. I need to check the difference.

gstreamer/src/date_time_serde.rs
fengalin: Oops, I will fix this.

gstreamer/src/format.rs
fengalin: Let me take another look at this.

gstreamer/src/buffer_serde.rs
sdroege: Could `Cow` be used for that?

gstreamer/src/format.rs
sdroege: No it's fine as-is, I just didn't think about the simple solution :)

gstreamer/src/segment_serde.rs
fengalin: I'm not quite sure. I must admit I don't have experience with the Formatted things so it's kind of fuzzy for me. Will investigate.

sdroege: Maybe there could be a simple macro implementation for those around the casts

gstreamer/src/value_serde.rs
fengalin: It could, but that would lead to things like `((1, 3))` due to the `Rational32` being an inner type. This also appears in some `FormattedValue`s, we might want to avoid these too (but that would lessen the use cases for my modification in `gir` :)).

gstreamer/src/format.rs
sdroege: Unless you think it leads to nicer serialization?

gstreamer/src/value_serde.rs
sdroege: I see :)

gstreamer/src/buffer_serde.rs
fengalin: Sure! Thanks.

gstreamer/src/format.rs
fengalin: Yes, I think we can do better than the `((things))` anyway.

gstreamer/src/buffer_serde.rs
sdroege: Can it really? That's great :)

gstreamer/src/caps_serde.rs
sdroege: You can only really check by `is_any()` and `is_empty()`, and then handle those cases separate from everything else somehow.

gstreamer/src/buffer_serde.rs
fengalin: After trying a bit, `Cow` doesn't seem to help. We would need to use either `Cow<'a, Bytes>` or `Cow<'a, ByteBuf>`. In the `Bytes` case, we need to copy the `[u8]` during deserialization because the `Deserializer` owns it and in the `ByteBuf` case, we need to copy before serialization due to its owning by nature. My proposal, although asymmetrical, avoids unneeded copy AFAICT.

gstreamer/src/caps_serde.rs
fengalin: Would that mean using a boolean (e.g. `is_any`) when there is no structures? We could use something like this (schematically):

```
caps: (
   is_any: bool,
   structures: [Structure],
)
```
but that would mean it could be inconsistent when `is_any` is `true` and `structures` is not empty.

gstreamer/src/date_time_serde.rs
fengalin: This raises additional questions. I'm heading toward this representation:

```
struct DateTimeSerde {
    y: Option<i32>,
    m: Option<i32>,
    time: Option<(i32, i32)>,
    s: Option<f64>,
    tz_offset: Option<f32>,
}
```

- I merged `h` and `mn` in `time` since `has_time` refers to both.
- I used an `Option` for `tz_offset` and I'm planning on using `new_local_time` in case it is `None`.
- They are no constructors for cases where `time` is defined, but `y`, `h` and `m` aren't, so I guess this would be the actual model to be used (schematically):
```
date_time: (
    y: Option<(
        i32,
        m: Option<(
            i32,
            time: Option<(
                i32,
                i32,
                s: Option<f64>,
                tz_offset: Option<f32>,
            )>,
        >,
    )>,
)
```

... but I don't really like it because I don't find it easy to read.

sdroege:
I'll get back to you about that tomorrow night when I'm at the computer again. But I agree generally.

For caps, maybe an `enum { Empty, Any, Something(&[StructureRef])`?

gstreamer/src/buffer_serde.rs
sdroege: I guess it could be an `Either<ByteBuf, Bytes>` but at that point it just becomes silly. Let's keep it as is then, thanks for checking :)

gstreamer/src/caps_serde.rs
sdroege: Proposal below to make it nicer https://github.com/sdroege/gstreamer-rs/pull/106#issuecomment-408617017 . Note that `Something` is just a placeholder, maybe you can find a better name :)

sdroege:
> Edit: there are options for this too. We can stay close to GStreamer naming with Any, Empty, ... or we can adopt a Rust-like approach: Any, None and Some(...). I'm in favor of the Rust-like approach, but here again, that could be confusing for people who are used to GStreamer.

`Any`, `Empty`, `Some(x)` seems nicest. I would prefer to keep the any and empty naming, because it's basically like sets and also because other functions are called empty/any and not none/any. It seems more consistent with the existing API.

gstreamer/src/date_time_serde.rs
fengalin: Yes, an `enum` seems better again. However, we might want to use abbreviations otherwise we get things like this:

```
YearMonthDayHourMinutesSecondsTzOffset(2018, 5, 28, 16, 6, 42841000, 2)
```

what do you think about this instead:

```
YMDHMnSTz(2018, 5, 28, 16, 6, 42841000, 2)
```

gstreamer/src/buffer_serde.rs
fengalin: I'll try it later, however since it wouldn't change the actual behaviour, I propose not to block the PR on this. I'll feel released somewhat when it is merged.

fengalin:
Sorry, one more thing for which I'm hesitating. I was somewhat influenced by the `GStreamer` serialization for `Values`. Currently, we get things like this:

```
my_array: [
    ("Fraction", (1, 3)),
    ("Fraction", (1, 2)),
    ("String", "test str"),
    ("u32", 42),
]
```

with an `enum` (again), we could probably use something like this:

```
my_array: [
    Fraction(1, 3),
    Fraction(1, 2),
    String("test str"),
    u32(42),
]
```

gstreamer/src/buffer_serde.rs
sdroege: Yes let's keep as-is and maybe add a TODO comment there for the future

gstreamer/src/date_time_serde.rs
sdroege: Sounds good to me

sdroege:
> with an enum (again), we could probably use something like this:

This would require adding all possible types there, but it's open-ended. IMHO not a good match for an enum but better stringly-typed

fengalin:
> This would require adding all possible types there, but it's open-ended. IMHO not a good match for an enum but better stringly-typed

I see. In that case, I think I have addressed all the points from our latest discussions.

gstreamer/src/date_time_serde.rs
```diff
@@ -7,53 +7,80 @@
...
+use serde::ser::{Serialize, Serializer};
 use DateTime;
 
+#[derive(Serialize, Deserialize)]
+enum DateTimeVariants {
+    Y(i32),
+    YM(i32, i32),
+    YMD(i32, i32, i32),
+    YMDhm(i32, i32, i32, i32, i32),
+    YMDhmsTz(i32, i32, i32, i32, i32, f64, f32),
```
sdroege: This seems not correct, from looking at the `GstDateTime` code:

```c
typedef enum
{
  GST_DATE_TIME_FIELDS_INVALID = 0,
  GST_DATE_TIME_FIELDS_Y,       /* have year                */
  GST_DATE_TIME_FIELDS_YM,      /* have year and month      */
  GST_DATE_TIME_FIELDS_YMD,     /* have year, month and day */
  GST_DATE_TIME_FIELDS_YMD_HM,
  GST_DATE_TIME_FIELDS_YMD_HMS
      /* Note: if we ever add more granularity here, e.g. for microsecs,
       * the compare function will need updating */
} GstDateTimeFields;
```

It seems that timezone is allowed regardless of everything else. So maybe we can add that simply as an `Option` into each variant?

gstreamer/src/tags.rs
```diff
@@ -39,7 +39,7 @@ macro_rules! impl_tag(
         }
 
         lazy_static! {
-            static ref $rust_tag: &'static str =
+            pub static ref $rust_tag: &'static str =
```
sdroege: This should probably be `pub(crate)`?

sdroege:
Can you double-check that none of those internal types and generally nothing else apart from the serde impls became public as part of this PR? Otherwise this indeed looks ready now :) Thanks for your patience

gstreamer/src/date_time_serde.rs
fengalin: Ok, this was not clear to me. With the `new_y`, `new_ym` & `new_ymd` there is no argument for the time zone. Should I use the `new` with all arguments version and assign `-1` to the missing parts? What should we do in case the time zone is not indicated? I know I said I would use the local time zone, but the available methods are kind of limited in that regard.

gstreamer/src/tags.rs
fengalin: It could. Note that this was part of #119 though. So we might want to address it as a separate PR (I can do it tonight).

fengalin:
Yes, I will. Thank you for your help.

gstreamer/src/date_time_serde.rs
sdroege: It uses `0.0` as timezone elsewhere for that

gstreamer/src/tags.rs
sdroege: OK, then let's do that :)

gstreamer/src/date_time_serde.rs
fengalin: Looking at `gst_date_time_get_time_zone_offset`, it seems that the time zone is only meaningful when `DateTime` holds time:

``` C
gst_date_time_get_time_zone_offset (const GstDateTime * datetime)
{
  g_return_val_if_fail (datetime != NULL, 0.0);
  g_return_val_if_fail (gst_date_time_has_time (datetime), 0.0);
  [...]
}
```

That would explain why the `new_y`, `new_ym` and `new_ymd` don't allow defining the time zone. If you confirm my understanding, I'll just add the `Option<f32>` to the `YMDhm` and `YMDhms` variants.

sdroege: Confirmed and agreed about no option

sdroege:
Looks all good to me. Would you like to squash/reorganize the commits a bit before merging? I leave that decision to you, I'm fine either way :)

fengalin:
I just merged the last `DateTime` fix with the initial `enum` implementation. It would not be very convenient to reorganize the rest now. Other than that, I must admit it's hard for me to squash considering the sweat it required :) Don't get me wrong, I enjoyed it and had the chance to try so many things I never tried before, including some of which that didn't make it to the final design like a `PhantomData` (which is really a phantom eventually).

Thanks again for your support and reactivity!

PR #113, opened 2018-06-11
Fix memory issue building a `Sample` with an `info` `Structure`

fengalin:
The following code in a `[test]`:
``` rust
let info = Structure::builder("sample.info")
    .field("f3", &123i32)
    .build();
let sample = Sample::new::<GenericFormattedValue>(
        None,
        None,
        None,
        Some(info.as_ref())
    );
```
generates one of the followings executing `cargo test`:
```
- segmentation fault
- signal: 6, SIGABRT: process abort signal
- signal: 11, SIGSEGV: invalid memory reference
```

Looking at [`gst_sample_new`](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstSample.html#gst-sample-new), it appears that the `GstStructure` argument `info` is `transfer full`, which makes sense since `GstStructure` "does not have a refcount because it usually is part of a higher level object" and "provides a means to enforce mutability using the refcount of the parent with the gst_structure_set_parent_refcount() method" ([source](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstStructure.html)).

So my best guess is to have `Sample::new` and `Sample::with_buffer_list` consume the `info` argument.

I also took the liberty to remove `mut_override` for the `segment` argument as it is declared as `const GstSegment *segment` in `gst_sample_new`. Or was it fixed from a previous version of `GStreamer`?

### DIFF
--- a/gstreamer/src/sample.rs
+++ b/gstreamer/src/sample.rs
@@ -22,6 +22,7 @@ use Caps;
 use FormattedSegment;
 use FormattedValue;
 use Segment;
+use Structure;
 use StructureRef;
 
 pub type Sample = GstRc<SampleRef>;
@@ -36,16 +37,16 @@ impl GstRc<SampleRef> {
         buffer: Option<&Buffer>,
         caps: Option<&Caps>,
         segment: Option<&FormattedSegment<F>>,
-        info: Option<&StructureRef>,
+        info: Option<Structure>,
     ) -> Self {
         assert_initialized_main_thread!();
         unsafe {
-            let info = info.map(|i| i.as_ptr()).unwrap_or(ptr::null());
+            let info = info.map(|i| i.into_ptr()).unwrap_or(ptr::null_mut());
 
             from_glib_full(ffi::gst_sample_new(
                 buffer.to_glib_none().0,
                 caps.to_glib_none().0,
-                mut_override(segment.to_glib_none().0),
+                segment.to_glib_none().0,
                 mut_override(info),
             ))
         }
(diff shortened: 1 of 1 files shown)


gstreamer/src/sample.rs
```diff
@@ -36,26 +38,29 @@ impl GstRc<SampleRef> {
         buffer: Option<&Buffer>,
         caps: Option<&Caps>,
         segment: Option<&FormattedSegment<F>>,
-        info: Option<&StructureRef>,
+        info: Option<Structure>,
     ) -> Self {
         assert_initialized_main_thread!();
         unsafe {
-            let info = info.map(|i| i.as_ptr()).unwrap_or(ptr::null());
+            let info_ptr = info.as_ref().map(|i| i.as_ptr()).unwrap_or(ptr::null());
```
sdroege: `i.into_ptr()` instead of `i.as_ptr()`, and then you can remove the `mem::forget` at the bottom :)

This is annoying, for `0.11` we should do a `info.to_glib_full()` here (which does a copy) for memory-safety.

fengalin: > i.into_ptr() instead of i.as_ptr(), and then you can remove the mem::forget at the bottom :)

Yes indeed, that's better!

> This is annoying, for 0.11 we should do a info.to_glib_full() here (which does a copy) for memory-safety.

Of course! I'll make another PR for 0.11.

sdroege: Thanks!

PR #114, opened 2018-06-11
0.11: Fix memory issue building a `Sample` with an `info` `Structure`

fengalin:
This is a workaround to get the fix from PR #113 on branch 0.11 without breaking the API.

I'm not too confident with `glib`, so I hope this is all right.

### DIFF
--- a/gstreamer/src/sample.rs
+++ b/gstreamer/src/sample.rs
@@ -40,13 +40,13 @@ impl GstRc<SampleRef> {
     ) -> Self {
         assert_initialized_main_thread!();
         unsafe {
-            let info = info.map(|i| i.as_ptr()).unwrap_or(ptr::null());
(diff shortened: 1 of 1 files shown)


gstreamer/src/structure.rs
```diff
@@ -505,6 +505,38 @@ impl PartialEq for StructureRef {
 
 impl Eq for StructureRef {}
 
+impl<'a> ToGlibPtr<'a, *const ffi::GstStructure> for StructureRef {
```
sdroege: Hmm we don't implement that for any of the other `Ref` types either. Maybe just directly call `gst_structure_copy()` up there instead of implementing all these traits. We're not going to use them anywhere else probably.

fengalin: And it's leaner. Thanks

PR #116, opened 2018-06-14
Change return type of add_probe to support removing

JohanGoversTech:


gstreamer/src/pad.rs
```diff
@@ -241,7 +241,11 @@ impl<O: IsA<Pad>> PadExtManual for O {
                 Some(destroy_closure),
             );
 
-            from_glib(id)
+            if id == 0 {
+                None
+            } else {
+                Some(PadProbeId(id))
```
sdroege: Please us `Some(from_glib(id))` here. Otherwise all fine, thanks a lot! :)

PR #119, opened 2018-07-08
Tags: implement generic iterators

fengalin:
This is the continuation of PR #107 by @ystreet

### DIFF
--- a/gstreamer/src/tags.rs
+++ b/gstreamer/src/tags.rs
@@ -188,6 +188,23 @@ impl TagListRef {
         }
     }
 
+    pub fn add_dynamic<'a, T>(&'a mut self, tag_name : &str, value: &T, mode: TagMergeMode)
+    where
+        T: ToSendValue
+    {
+        /* TODO: implement limiting values for tags? */
+        unsafe {
+            let v = value.to_send_value();
+
+            ffi::gst_tag_list_add_value(
+                self.as_mut_ptr(),
+                mode.to_glib(),
+                tag_name.to_glib_none().0,
+                v.to_glib_none().0,
+            );
+        }
+    }
+
     pub fn get<'a, T: Tag<'a>>(&self) -> Option<TypedValue<T::TagType>> {
         unsafe {
             let mut value: Value = mem::zeroed();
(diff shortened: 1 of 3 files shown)


sdroege:
> This is the continuation of PR #107 by @ystreet

If it's based on his PR, please include his commits verbatim here in this PR and add your changes on top in separate commits :)

gstreamer/src/tags.rs
```diff
@@ -477,6 +237,32 @@ impl TagListRef {
         }
     }
 
+    pub fn get_generic(&self, tag_name: &str) -> Option<Value> {
```
sdroege: Should probably be a `SendValue` for symmetry

gstreamer/src/tags.rs
```diff
@@ -493,14 +279,42 @@ impl TagListRef {
         }
     }
 
-    pub fn get_size<'a, T: Tag<'a>>(&'a self) -> u32 {
+    pub fn get_index_generic<'a>(&'a self, tag_name: &str, idx: u32) -> Option<&'a Value> {
```
sdroege: And also here `SendValue`

Would it be possible to implement the old typed getters/setters around the new generic ones now to get rid of some unsafe code? That would be great

gstreamer/src/tags.rs
```diff
@@ -627,6 +441,127 @@ where
...
+            taglist: taglist,
+            name,
+            idx: 0,
+            size: taglist.get_size_by_name(name),
+        }
+    }
+}
+
+impl<'a> Iterator for GenericTagIterator<'a> {
+    type Item = &'a Value;
```
sdroege: Also `SendValue` I guess

sdroege:
Generally looks good, thanks!

fengalin:
I couldn't find a better way for the `get_index` factorization. Is it what you were expecting?

gstreamer/src/tags.rs
```diff
@@ -264,19 +264,15 @@ impl TagListRef {
...
-            Some(&*(value as *const TypedValue<T::TagType>))
-        }
+        self.get_index_generic(T::tag_name(), idx)
+            .and_then(|value| unsafe {
+                let value = value.to_glib_none().0;
+                if (*value).g_type != T::TagType::static_type().to_glib() {
+                    return None;
+                }
+
+                Some(&*(value as *const TypedValue<T::TagType>))
```
sdroege: Why is the `unsafe` here needed? It could a) simply assert if types don't match and there's b) [a way for downcasting](https://github.com/gtk-rs/glib/blob/b8b98c9648326b399f5059f89746d440b8d6f05d/src/value.rs#L641-L643) with safe code that basically does the same you are doing here :)

But otherwise yes, that's what I was thinking of. And the same could possibly be done for the un-indexed getter and the setter/adder function?

fengalin: >  b) a way for downcasting with safe code that basically does the same you are doing here :)

I struggled trying to get that working, but had issues with `downcast` not returning a reference and the `Tag`s types not implementing `FromValueOptional` (can't be more precise right now, sorry). I will try again tonight or later this week.

sdroege: Ok thanks, put your non-working code in a branch then and I'll take a look :)

fengalin: Here's a commit in a dedicated branch with the failing code: https://github.com/fengalin/gstreamer-rs/tree/factorize_tags_fn

AFAICT, `downcast` consumes the `SendValue` and returns an owned `TypedValue` when we get an `Option<&'a SendValue>` and want an `Option<&'a TypedValue<T::TagType>>`.

sdroege: Oh I see. I guess we want a `downcast_ref` for all those variants in `value.rs` in glib :)

sdroege: Let me just add that :)

sdroege: https://github.com/gtk-rs/glib/pull/356

fengalin: Awesome! Thanks! I'm proceeding with the changes in `tags` then.

fengalin: Just one thing, it seems to me there is a mismatch between the comment "Returns `Ok(T)` if the object is an instance of `T` and `Err(self)` otherwise" and the actual signature.

sdroege: Thanks, fixed

sdroege:
Thanks, this looks ready now :)

PR #124, opened 2018-08-04
gstreamer: Implement deinit() function

philn:
This function is especially usefull when using the leaks tracer. It was removed
in commit e7a0543c.

### DIFF
--- a/gstreamer/src/lib.rs
+++ b/gstreamer/src/lib.rs
@@ -255,6 +255,10 @@ pub fn init() -> Result<(), glib::Error> {
     }
 }
 
+pub unsafe fn deinit() {
+    ffi::gst_deinit();
+}
+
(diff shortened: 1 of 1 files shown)


sdroege:
This function is unsafe. Calling any GStreamer function afterwards can misbehave in any possible way, including memory safety issues.

I'm fine with having it in the bindings as unsafe function though.

philn:
Like this? ^^ :)

philn:
Yeah, my GstPlayer has a refcount of 28 when `deinit()` is called. My app needs quite a lot of yak-shaving :/

sdroege:
The gtk/gstreamer example might give some ideas. But yeah, I also didn't consider reference cycles seriously enough in many places

PR #128, opened 2018-08-18
examples: Explicitly define [bin] section for discoverer

ford-prefect:


sdroege:
Thanks, I'll merge that once gdk builds again and the CI is happy again

PR #132, opened 2018-10-08
Gitlab CI config

alatiera:
Preparing for the gitlab migration

One thing to note is that travis was using a `ubuntu trusty` base, where now we are using the rust docker images, both based on debian stable. Sadly the tarball hack is still with us, though I plan on getting rid of it in the future.

There is also a `rustfmt` job, currently the pipeline becomes red if it fails, but it can also be set to `allow_failure: true`.

Pipeline results: https://gitlab.freedesktop.org/alatiera/gstreamer-rs/pipelines/5458

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,51 @@
+stages:
+  - "test"
+
+.cargo_test_template: &cargo_test
+  stage: "test"
+  variables:
+    DEPENDENCIES: |
+      curl
+      liborc-dev
+      libglib2.0-dev
+      libxml2-dev
+      libgtk-3-dev
+      libegl1-mesa
+      libgles2-mesa
+      libgl1-mesa-dri
+      libgl1-mesa-glx
+      libwayland-egl1-mesa
+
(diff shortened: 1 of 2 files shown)


sdroege:
Thanks!

travis was using Ubuntu trusty because that's the newest they provide. Anything else is an improvement already :)

sdroege:
The travis configuration also builds (and runs tests) with --all-features btw, and checks with stable, *beta* and nightly to make sure it all stays working.

The beta is especially helpful... gst-plugin-rs currently fails on beta but not on stable and nightly. There's a compiler bug in beta :)

alatiera:
> The travis configuration also builds (and runs tests) with --all-features btw

gitlab does the exact same, literally copied the script from travis and added --color=always.

> The beta is especially helpful... gst-plugin-rs currently fails on beta but not on stable and nightly. There's a compiler bug in beta :)

There's no pre-built image for the beta channel, but we *could* build one. It doesn't take much maintenance, I am already doing it for `librsvg`.

PR #138, opened 2018-10-19
Update gir

ceyusa:
This change needs https://github.com/sdroege/gstreamer-sys/pull/18


gstreamer-audio/src/audio_format_info.rs
```diff
@@ -126,7 +126,7 @@ impl AudioFormatInfo {
                 self.0,
                 flags.to_glib(),
                 dest.as_mut_ptr() as *mut _,
-                src.as_ptr() as *mut _,
+                src.as_ptr() as _,
```
sdroege: Is a cast needed at all here now?

ceyusa: afaiu yes, because u8 is not &[u8]

sdroege: See https://github.com/sdroege/gstreamer-sys/pull/18/files#r226812757

ceyusa:
fixed, remove the unneeded patch for audio and a quick fix for README.md

PR #142, opened 2018-10-28
Cturner/add tutorial 9

charlie-ht:
The build is not working correctly yet, posting this PR for some advice on that :-)

```
[cturner@localhost tutorials]$ cargo build
    Updating git repository `https://github.com/sdroege/gstreamer-rs`
    Updating git repository `https://github.com/sdroege/gstreamer-sys`                                                         
    Updating crates.io index                                                                                                   
error: no matching version `= 0.5.0` found for package `gstreamer-audio-sys`                                                   
location searched: https://github.com/sdroege/gstreamer-sys#d68a12a4
versions found: 0.7.0
required by package `gstreamer-pbutils-sys v0.7.0 (https://github.com/sdroege/gstreamer-sys#a4113052)`
    ... which is depended on by `gstreamer-pbutils v0.13.0 (https://github.com/sdroege/gstreamer-rs#8a6bcbce)`
    ... which is depended on by `tutorials v0.11.0 (/home/cturner/git/gstreamer-rs/tutorials)`
```

It wasn't clear to me how I should depend on gstreamer-pbutils in the tutorials' `Cargo.toml` file.  There is also an issue with printing caps in this patch, I have a description of the issue in the code, maybe there is a nicer way to handle printing GValues.

### DIFF
--- a/tutorials/Cargo.toml
+++ b/tutorials/Cargo.toml
@@ -11,7 +11,9 @@ gstreamer = { path = "../gstreamer" }
 gstreamer-audio = { path = "../gstreamer-audio" }
 gstreamer-video = { path = "../gstreamer-video" }
 gstreamer-app = { path = "../gstreamer-app" }
+gstreamer-pbutils = { path = "../gstreamer-pbutils" }
 byte-slice-cast = "0.2"
+failure = "0.1.0"
 
 [badges]
 travis-ci = { repository = "sdroege/gstreamer-rs", branch = "master" }
--- a/tutorials/src/bin/basic-tutorial-9.rs
+++ b/tutorials/src/bin/basic-tutorial-9.rs
@@ -1,0 +1,184 @@
+extern crate failure;
+use failure::Error;
+
+extern crate glib;
+extern crate gstreamer as gst;
+extern crate gstreamer_pbutils as gst_pbutils;
+use gst_pbutils::{
+    prelude::*, Discoverer, DiscovererContainerInfo, DiscovererInfo, DiscovererResult,
+    DiscovererStreamInfo,
+};
+
+use std::env;
+
+#[path = "../tutorials-common.rs"]
+mod tutorials_common;
+
+fn send_value_as_str(v: &glib::SendValue) -> Option<String> {
+    if let Some(s) = v.get::<&str>() {
+        Some(s.to_string())
+    } else if let Some(serialized) = v.serialize() {
+        Some(serialized)
+    } else {
+        None
+    }
+}
+
+fn print_stream_info(info: &DiscovererStreamInfo, depth: usize) {
+    let caps_str = if let Some(caps) = info.get_caps() {
+        if caps.is_fixed() {
+            gst_pbutils::pb_utils_get_codec_description(&caps)
+                .unwrap_or_else(|| String::from("unknown codec"))
+        } else {
+            caps.to_string()
+        }
+    } else {
+        String::new()
+    };
+
+    let stream_nick = info.get_stream_type_nick();
+    println!(
+        "{stream_nick:>indent$}: {caps_str}",
+        stream_nick = stream_nick,
+        indent = 2 * depth + stream_nick.len(),
+        caps_str = caps_str
+    );
+
+    if let Some(tags) = info.get_tags() {
+        println!("{:indent$}Tags:", " ", indent = 2 * depth);
+        for (tag, values) in tags.iter_generic() {
+            let mut tags_str = format!(
+                "{tag:>indent$}: ",
+                tag = tag,
+                indent = 2 * (2 + depth) + tag.len()
+            );
+            // reviewer: I'm intending to write something along the lines of
+            // println!("{}", values.iter().collect(send_value_as_str).join(",")
+            let mut tag_num = 0;
+            for value in values {
+                if let Some(s) = send_value_as_str(value) {
+                    if tag_num > 0 {
+                        tags_str.push_str(", ")
+                    }
+                    tags_str.push_str(&s[..]);
+                    tag_num += 1;
+                }
+            }
+
+            println!("{}", tags_str);
+        }
+    };
+}
+
+/* Print information regarding a stream and its substreams, if any */
+fn print_topology(info: &DiscovererStreamInfo, depth: usize) {
+    print_stream_info(info, depth);
+
+    if let Some(next) = info.get_next() {
+        print_topology(&next, depth + 1);
+    } else if let Some(container_info) = info.downcast_ref::<DiscovererContainerInfo>() {
+        for stream in container_info.get_streams() {
+            print_topology(&stream, depth + 1);
+        }
+    }
+}
+
+fn on_connect_discovered(
+    _discoverer: &Discoverer,
+    discoverer_info: &DiscovererInfo,
+    error: &Option<glib::Error>,
+) {
+    let uri = discoverer_info.get_uri().unwrap();
+    match discoverer_info.get_result() {
+        DiscovererResult::Ok => println!("Discovered {}", uri),
+        DiscovererResult::UriInvalid => println!("Invalid uri {}", uri),
+        DiscovererResult::Error => if let Some(msg) = error {
+            println!("{}", msg);
+        } else {
+            println!("Unknown error")
+        },
+        DiscovererResult::Timeout => println!("Timeout"),
+        DiscovererResult::Busy => println!("Busy"),
+        DiscovererResult::MissingPlugins => if let Some(s) = discoverer_info.get_misc() {
+            println!("{}", s);
+        },
+        _ => println!("Unknown result"),
+    }
+
+    if discoverer_info.get_result() != DiscovererResult::Ok {
+        return;
+    }
+
+    println!("Duration: {}", discoverer_info.get_duration());
+
+    if let Some(tags) = discoverer_info.get_tags() {
+        println!("Tags:");
+        for (tag, values) in tags.iter_generic() {
+            print!("  {}: ", tag);
+            values.for_each(|v| {
+                if let Some(s) = send_value_as_str(v) {
+                    println!("{}", s)
+                }
+            })
+        }
+    }
+
+    println!(
+        "Seekable: {}",
+        if discoverer_info.get_seekable() {
+            "yes"
+        } else {
+            "no"
+        }
+    );
+
+    println!("Stream information:");
+
+    if let Some(stream_info) = discoverer_info.get_stream_info() {
+        print_topology(&stream_info, 1);
+    }
+}
+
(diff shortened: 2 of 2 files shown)


tutorials/Cargo.toml
```diff
@@ -11,8 +11,10 @@ gstreamer = { path = "../gstreamer" }
 gstreamer-audio = { path = "../gstreamer-audio" }
 gstreamer-video = { path = "../gstreamer-video" }
 gstreamer-app = { path = "../gstreamer-app" }
+gstreamer-pbutils = { git = "https://github.com/sdroege/gstreamer-rs" }
```
sdroege: You would provide the path here, like in the lines above :)

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,150 @@
+extern crate failure;
+use failure::Error;
+
+extern crate glib;
+extern crate gstreamer as gst;
+extern crate gstreamer_pbutils as pbutils;
```
sdroege: Elsewhere we keep the name, `gst_pbutils`

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,150 @@
...
+        let desc = if caps.is_fixed() {
+            pbutils::pb_utils_get_codec_description(&caps)
+                .unwrap_or_else(|| String::from("unknown codec"))
+        } else {
+            caps.to_string()
+        };
+        println!(" {}", desc);
+    } else {
+        println!();
+    }
```
sdroege: This would probably be nicer with building a single string and then printing it all at once. E.g.

```rust
let caps_str = if let Some(caps) ... {
    ...
} else {
    String::new()
};

println!(....);
```

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,150 @@
...
+        println!("{:indent_level$}Tags:", " ", indent_level = 2 * depth);
+        tags.iter_tag_list().for_each(|(tag, values)| {
+            print!("{:indent_level$}", " ", indent_level = 2 * (2 + depth));
+            print!("{}: ", tag);
+            values.for_each(|v| {
+                // reviewer: this isn't nice, but if you call
+                // serialize() on a string it escapes it since it has
+                // to be able to read it back again, which makes for
+                // ugly looking output. Is there a better way?
+                if let Some(s) = v.downcast_ref::<&str>() {
```
sdroege:
```rust
if let Some(s) = v.get::<&str>() { ... }
```

maybe?

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,150 @@
...
+        println!(" {}", desc);
+    } else {
+        println!();
+    }
+
+    if let Some(tags) = info.get_tags() {
+        println!("{:indent_level$}Tags:", " ", indent_level = 2 * depth);
+        tags.iter_tag_list().for_each(|(tag, values)| {
+            print!("{:indent_level$}", " ", indent_level = 2 * (2 + depth));
+            print!("{}: ", tag);
```
sdroege: Also here I prefer to do a single print per line :)

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,150 @@
...
+            caps.to_string()
+        };
+        println!(" {}", desc);
+    } else {
+        println!();
+    }
+
+    if let Some(tags) = info.get_tags() {
+        println!("{:indent_level$}Tags:", " ", indent_level = 2 * depth);
+        tags.iter_tag_list().for_each(|(tag, values)| {
```
sdroege: Instead of the `for_each`, often it is nicer to do

```rust
for (tag, values) in tags.iter_tag_list() {
    ....
}
```

You could also use `tags.iter()` now, which gives you a merged value for each tag.

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,150 @@
...
+    discoverer_info: &DiscovererInfo,
+    error: &Option<glib::Error>,
+) {
+    let uri = discoverer_info.get_uri().unwrap();
+    match discoverer_info.get_result() {
+        DiscovererResult::Ok => println!("Discovered {}", uri),
+        DiscovererResult::UriInvalid => println!("Invalid uri {}", uri),
+        DiscovererResult::Error => if let Some(msg) = error {
+            println!("{}", msg);
+        },
```
sdroege: else unknown error?

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    if discoverer_info.get_result() != DiscovererResult::Ok {
+        return;
+    }
+
+    println!("Duration: {}", discoverer_info.get_duration());
+
+    if let Some(tags) = discoverer_info.get_tags() {
+        println!("Tags:");
+        tags.iter_tag_list().for_each(|(tag, values)| {
```
sdroege: Maybe `for (tag, value) in tags.iter() { ... }`

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,150 @@
...
+    if let Some(stream_info) = discoverer_info.get_stream_info() {
+        print_topology(&stream_info, 1);
+    }
+}
+
+fn main() -> Result<(), Error> {
+    gst::init()?;
+
+    let args: Vec<_> = env::args().collect();
+    let uri: &str = if args.len() == 2 {
```
sdroege: `>= 2` maybe?

sdroege: Maybe we also just want discoverer all args async instead of just the first?

tutorials/Cargo.toml
sdroege: `../gstreamer-pbutils`

sdroege:
There's also a merge conflict with `Cargo.toml`

tutorials/Cargo.toml
charlie-ht: Hah, ok. Is `cargo build` in the top-level dir supposed to work? I'm hitting what appears to be an unrelated error while testing the toml patch,

```
[cturner@localhost gstreamer-rs]$ pwd
/home/cturner/git/gstreamer-rs
[cturner@localhost gstreamer-rs]$ cargo build
    Updating crates.io index
    Updating git repository `https://github.com/gtk-rs/sys`
    Updating git repository `https://github.com/sdroege/gstreamer-sys`                                                         
    Updating git repository `https://github.com/gtk-rs/pangocairo`                                                             
error: no matching version `^0.5.0` found for package `glib-sys`                                                               
location searched: https://github.com/gtk-rs/sys
versions found: 0.7.0
required by package `glib v0.4.0 (https://github.com/gtk-rs/glib#a78f961b)`
    ... which is depended on by `gstreamer v0.13.0 (/home/cturner/git/gstreamer-rs/gstreamer)`
```

sdroege: `cargo build --all` and run a `cargo update` after you changed dependencies

sdroege:
@charlie-ht please ping me when I should review again, or write to me on IRC if you need some more guidance :)

tutorials/src/bin/basic-tutorial-9.rs
charlie-ht: That would be different behaviour to the C tutorial, although arguably a little more interesting. One problem is that I haven't read the concurrency parts of rust which would be required to keep an atomic counter of the remaining uri's to be discovered so that loop.quit() can be properly called once all discoveries have been made. The C API for discovery doesn't appear to give you any calls to check if the loop is idle, so this would need to be added to the Rust code, perhaps complicating it more than necessary to demonstrate usage of the discovery API.

charlie-ht:
> @charlie-ht please ping me when I should review again, or write to me on IRC if you need some more guidance :)

Thanks @sdroege :)

tutorials/src/bin/basic-tutorial-9.rs
sdroege: Good point. Let's keep it simple then :)

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,184 @@
...
+    if let Some(tags) = info.get_tags() {
+        println!("{:indent$}Tags:", " ", indent = 2 * depth);
+        for (tag, values) in tags.iter_generic() {
+            let mut tags_str = format!(
+                "{tag:>indent$}: ",
+                tag = tag,
+                indent = 2 * (2 + depth) + tag.len()
+            );
+            // reviewer: I'm intending to write something along the lines of
+            // println!("{}", values.iter().collect(send_value_as_str).join(",")
```
sdroege: That would allocate a temporary vector though :) https://docs.rs/itertools/0.7.8/itertools/trait.Itertools.html#method.join might be an option here, or you can do it with `Iterator::fold()` but the way you do it is also fine

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,184 @@
...
+    let uri: &str = if args.len() == 2 {
+        args[1].as_ref()
+    } else {
+        "https://www.freedesktop.org/software/gstreamer-sdk/data/media/sintel_trailer-480p.webm"
+    };
+
+    println!("Discovering {}", uri);
+
+    let loop_ = glib::MainLoop::new(None, false);
+    let timeout = gst::ClockTime::from_seconds(5);
```
sdroege: This could also be `5 * gst::SECOND` :)

tutorials/src/bin/basic-tutorial-9.rs
```diff
@@ -0,0 +1,184 @@
...
+    discoverer.discover_uri_async(uri)?;
+
+    loop_.run();
+
+    discoverer.stop();
+
+    Ok(())
+}
+
+fn tutorial_main() {
```
sdroege: You could make this return a `Result<(), Error>`, and the same for `main()`

tutorials/src/bin/basic-tutorial-9.rs
charlie-ht: For this code I was more interested in readability than performance, but since it's OK as it is, I'll leave it be :)

tutorials/src/bin/basic-tutorial-9.rs
charlie-ht: I needed some indirection because having the trailing return type on main was making the compiler upset, since it appears `tutorials_common::run` doesn't like the return value.

PR #145, opened 2018-10-29
Use FlowError instead of FlowReturn as error part of Result

ford-prefect:
This prevents the possibility of a Result that is an Err(Ok), which
would not be meaningful.

### DIFF
--- a/gstreamer/src/enums.rs
+++ b/gstreamer/src/enums.rs
@@ -95,6 +95,13 @@ impl FlowReturn {
         }
     }
 
+    pub fn into_result_value<T, F: FnOnce() -> T>(self, func: F) -> Result<T, FlowError> {
+        match self.into_result() {
+            Ok(_) => Ok(func()),
+            Err(err) => Err(err),
+        }
+    }
+
     pub fn from_error(v: FlowError) -> Self {
         match v {
             FlowError::NotLinked => FlowReturn::NotLinked,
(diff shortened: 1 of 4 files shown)


sdroege:
Interesting how the test fails. Any ideas?

ford-prefect:
D'oh, the argument of into_result_success() can't work this way -- it needs
to be lazily evaluated. Need to figure out how to do that, or maybe change
that function a bit.

sdroege:
Like or_else on Option. Closures ftw :)

ford-prefect:
I did it with Closures, but I'm not sure this is idiomatic?

sdroege:
> I did it with Closures, but I'm not sure this is idiomatic?

That's exactly how it's done everywhere :) Thanks!

gstreamer/src/enums.rs
```diff
@@ -95,6 +95,13 @@ impl FlowReturn {
         }
     }
 
+    pub fn into_result_value<T, F: Fn() -> T>(self, func: F) -> Result<T, FlowError> {
```
sdroege: This should be an `FnOnce` as that allows you to do much more in the closure. See signature of `Option::unwrap_or_else()` etc.

ford-prefect:
Is it normal to have do `into_result_value(|| ...)` instead of `into_result_value(...)`? I know `unwrap_or_else()` doesn't need the explicit definition of a closure (otoh it's a bit better to have it explicit, maybe).

gstreamer/src/enums.rs
ford-prefect: Is it normal to have do `into_result_value(|| ...)` instead of `into_result_value(...)`? I know `unwrap_or_else()` doesn't need the explicit definition of a closure (otoh it's a bit better to have it explicit, maybe).

sdroege: It also needs the closure. That's why there is unwrap_or and unwrap_or_else

PR #150, opened 2018-11-02
Generate bindings for the GStreamer Editing Services

thiblahute:


thiblahute:
Thanks for all the help Sebastian!

examples/src/bin/ges.rs
```diff
@@ -0,0 +1,94 @@
...
+#[path = "../examples-common.rs"]
+mod examples_common;
+
+extern crate failure;
+#[allow(unused_imports)]
+use failure::Error;
+
+fn main_loop(uri: &str) -> Result<(), Error> {
+    gst::init()?;
+    ges::init().unwrap();
```
MathieuDuponchelle: can't you get rid of these unwraps?

sdroege: You could simply add another ? here instead :)

gstreamer-editing-services/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+#[macro_use]
+extern crate bitflags;
+
+// Re-export all the traits in a prelude module, so that applications
+// can always "use gst::prelude::*" without getting conflicts
+pub mod prelude {
+    pub use glib::prelude::*;
+    pub use gst::prelude::*;
+
+    pub use auto::traits::*;
```
sdroege: You want to re-export `TimelineElementExtManual` here

gstreamer-editing-services/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+
+pub use glib::{Cast, Continue, Error, IsA, StaticType, ToValue, Type, TypedValue, Value};
+
+#[cfg_attr(feature = "cargo-clippy", allow(unreadable_literal))]
+#[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
+#[cfg_attr(feature = "cargo-clippy", allow(too_many_arguments))]
+#[cfg_attr(feature = "cargo-clippy", allow(match_same_arms))]
+mod auto;
+pub use auto::*;
+
```
sdroege: `mod timeline_element;`

gstreamer-editing-services/src/timeline_element.rs
```diff
@@ -0,0 +1,67 @@
...
+                self.to_glib_none().0,
+                name.to_glib_none().0,
+                ptr::null_mut(),
+                ptr::null_mut(),
+            ));
+            if !found {
+                return None;
+            }
+
+            letPaPad value = glib::Value::uninitialized();
```
sdroege: How does this compile? ;)

gstreamer-editing-services/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+#[macro_use]
+extern crate glib;
+extern crate gio;
+
+static GES_INIT: Once = ONCE_INIT;
+
+pub fn init() -> Result<(), &'static str> {
+    unsafe {
+        if from_glib(ffi::ges_init()) {
```
sdroege: You might want to also first call `gst::init()` here?

gstreamer-editing-services/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+            Ok(())
+        } else {
+            Err("Could not initialize GES.")
```
sdroege: `glib::BoolError` instead of a plain string

gstreamer-editing-services/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+macro_rules! assert_initialized_main_thread {
+    () => {
+        if unsafe { ::gst_ffi::gst_is_initialized() } != ::glib_ffi::GTRUE {
+            panic!("GStreamer has not been initialized. Call `gst::init` first.");
+        }
+        ::GES_INIT.call_once(|| {
+            unsafe { ::ffi::ges_init() };
+        });
```
sdroege: Do we want lazy initialization of ges if there is also an explicit init function?

Gir_GstEditingServices.toml
```diff
@@ -0,0 +1,111 @@
+[options]
+girs_dir = "gir-files"
+library = "GES"
+version = "1.0"
+min_cfg_version = "1.14"
+target_path = "gstreamer-editing-services"
+work_mode = "normal"
+# concurrency = "send+sync"
```
sdroege: This should probably be enabled?

gstreamer-editing-services/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+version = "0.13.0"
+authors = ["Thibault Saunier <tsaunier@igalia.com>", "Sebastian Dröge <sebastian@centricular.com>"]
+categories = ["api-bindings", "multimedia"]
+description = "Rust bindings for GStreamer Editing Services"
+repository = "https://github.com/sdroege/gstreamer-rs"
+license = "MIT/Apache-2.0"
+readme = "README.md"
+homepage = "https://gstreamer.freedesktop.org"
+documentation = "https://sdroege.github.io/rustdoc/gstreamer/gstreamer_editing_services"
+keywords = ["gstreamer", "multimedia", "audio", "video", "gnome", "video editing", "nle"]
```
sdroege: crates.io only allows up to 5 keywords :)

sdroege:
Looks good to me apart from those few comments :)

gstreamer-editing-services/src/timeline_element.rs
thiblahute: Erg... forgot to add that file again :-) ... should be fixed.

gstreamer-editing-services/src/lib.rs
thiblahute: Right!

gstreamer-editing-services/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+#[cfg_attr(feature = "cargo-clippy", allow(match_same_arms))]
+mod auto;
+pub use auto::*;
+
+mod timeline_element;
+
+#[macro_use]
+extern crate bitflags;
+
+pub use timeline_element::TimelineElementExtManual;
```
sdroege: Move this up to below the `mod timeline_element;` line

gstreamer-editing-services/src/lib.rs
sdroege: `glib::BoolError::from_glib(something_that_gives_gboolean, "Some string")`

gstreamer-editing-services/src/lib.rs
thiblahute: Sorry, it was done locally :|

gstreamer-editing-services/src/lib.rs
thiblahute: Same :|

gstreamer-editing-services/src/lib.rs
thiblahute: I am not sure, do you think it is better?

sdroege: Probably not. But is explicit initialization of GES really needed or can we just do it lazily?

gstreamer-editing-services/src/lib.rs
sdroege: If it can be done lazily and never fails anyway, sure

examples/src/bin/ges.rs
thiblahute: Ended up getting read of then :-)

gstreamer-editing-services/src/lib.rs
thiblahute: `Err(BoolError("Could not initialize GES."))` seems to work, isn't it correct?

gstreamer-editing-services/src/lib.rs
thiblahute: It can, if user want to set some option he needs to do it manually, sounds safe to me.

gstreamer-editing-services/src/lib.rs
sdroege: That also works, yes.

Gir_GstEditingServices.toml
sdroege: So everything but the `GstElement` subclasses in GES are not actually thread-safe, so the above is correct but there's still a problem.

First of all, you should move the above configuration to all `GstElement` subclasses (or other `GstObject` subclasses) for consistency.

The actual problem however is that you can add a `GESLayer` to a `GESTimeline` (a `GstElement), and then get that very `GESLayer` out of the `GESTimeline` from another thread. But as `GESLayer` is not thread-safe, this would have to require unsafe code, which it doesn't currently. That's the whole point of the Rust thread-safety guarantees.

Note that this is unrelated of configuring `send+sync` on `GESTimeline`: you could always downcast to `GstElement`, send to a different thread and then upcast to `GESTimeline` again.

I'm not sure what to do about this here, this is a real problem for Rust.

sdroege:
This is now tracked at https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/merge_requests/150

PR #151, opened 2018-11-15
fix initialization of ptp clock

cmeissl:
The ptp clock was initialized as an ntp clock.


sdroege:
Thanks! gstreamer-rs moved to gitlab but I've imported your changes as a MR: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/merge_requests/169

PR #158, opened 2019-12-16
Bindings for AudioSink

hatzel:
All this does is provide rust bindings for the audio source so that it's convenience/features (over BaseSink) can be used.

I am sure these are not perfect but it seems to work well with the closed source sink I have written.

The one thing that irks me the most is that no methods of `BaseSink` can not be overwritten in structs based on AudioSink. Otherwise the render method would have to be implemented, resulting in `write` not being called.

One thing we might want to consider is making write mandatory to implement (i.e. not providing a default implementation).


sdroege:
Thanks! The correct place for merge requests is https://gitlab.freedesktop.org/gstreamer/gstreamer-rs

This repository here is only for historical reasons, everything was moved into the GStreamer gitlab more than a year ago.

PR #161, opened 2020-04-01
examples: remove the use of description() in Error

cotigao:
Replace the use of deprecated method "Error::description()"
with to_string() method


sdroege:
Thanks, apparently I missed those in https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/commit/ca6bcf36ab59ce7d26310bdc6d29dea45e9ca250 :)

Note that the correct repository for this is at https://gitlab.freedesktop.org/gstreamer/gstreamer-rs but I've merged your changes there now.

PR #164, opened 2020-06-05
Add remark about pkg-config in MSYS2

jneem:
Mention (and link to) the known issue about the MSYS2 version of pkg-config.


sdroege:
Thanks, I've merged this and added a remark about https://sourceforge.net/projects/pkgconfiglite/.

This is the wrong repository though, it's only a mirror of the real one. See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/commit/e7cb758ee31770a455d21e8bd59c2df83bb9c1fd :)

PR #179, opened 2021-12-13
Add the MessageRef::has_name() utility

pbor:


gstreamer/src/message.rs
```diff
@@ -68,6 +68,11 @@ impl MessageRef {
         }
     }
 
+    #[doc(alias = "gst_message_has_name")]
+    pub fn has_name(&self, name: &str) -> bool {
+        self.structure.map_or(false, |s| s.name() == name)
```
sdroege, line 73: You could add the same for structure

sdroege:
https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/930

PR #199, opened 2025-01-09
Fix Buffer Serialize struct length

jeremy-prater:
I noticed that `rmp-serde` and `serde_cbor`/`ciborium` both fail to deserialize the gstreamer::Buffer object because the encoded struct size is incorrect. There are 7 elements in the struct, but the serializer only specifies 6.

It is curious that `serde_json` worked correctly, but binary serialization did not.


sdroege:
Thanks. This is only a mirror of the official repository. Can you create a MR here please: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs ?

